PR #7, opened 2024-10-03
Add comments by Jon

joncinque:
I finally looked through the program, and it looks great! There are a few footguns in general just due to the native -> BPF differences, but overall everything looks good. I even compared the helpers line by line to make sure they were all doing the same thing.

Let me know what you think

### DIFF
--- a/program/src/processor.rs
+++ b/program/src/processor.rs
@@ -29,12 +29,16 @@ use {
 // its a gettable sysvar ie, allow people to omit them from the accounts list
 // without breaking compat to be done after release, we keep the existing
 // interface for all instructions for compat with firedancer
+// JC: agreed!
 
 fn get_vote_state(vote_account_info: &AccountInfo) -> Result<Box<VoteState>, ProgramError> {
     if *vote_account_info.owner != solana_vote_program::id() {
         return Err(ProgramError::IncorrectProgramId);
     }
 
+    // JC: I think the code was added after this commit, but can we switch to
+    // `deserialize_into_uninit` to avoid the overhead of creating a
+    // `VoteState::default`?
     let mut vote_state = Box::<VoteState>::default();
     VoteState::deserialize_into(&vote_account_info.data.borrow(), &mut vote_state)
         .map_err(|_| ProgramError::InvalidAccountData)?;
@@ -69,15 +73,25 @@ fn relocate_lamports(
     destination_account_info: &AccountInfo,
     lamports: u64,
 ) -> ProgramResult {
-    let mut source_lamports = source_account_info.try_borrow_mut_lamports()?;
-    **source_lamports = source_lamports
-        .checked_sub(lamports)
-        .ok_or(ProgramError::InsufficientFunds)?;
+    // JC: can we put this in separate blocks to make sure that we don't have a
+    // double mutable borrow? Along with a comment like this:
+    //
+    // If source and destination are the same account, there's a possibility to
+    // have two mutable borrows at the same time. To avoid that case, we use
+    // two separate blocks and ensure that the mutable borrow is dropped.
+    {
+        let mut source_lamports = source_account_info.try_borrow_mut_lamports()?;
+        **source_lamports = source_lamports
+            .checked_sub(lamports)
+            .ok_or(ProgramError::InsufficientFunds)?;
+    }
 
-    let mut destination_lamports = destination_account_info.try_borrow_mut_lamports()?;
-    **destination_lamports = destination_lamports
-        .checked_add(lamports)
-        .ok_or(ProgramError::ArithmeticOverflow)?;
+    {
+        let mut destination_lamports = destination_account_info.try_borrow_mut_lamports()?;
+        **destination_lamports = destination_lamports
+            .checked_add(lamports)
+            .ok_or(ProgramError::ArithmeticOverflow)?;
+    }
 
     Ok(())
 }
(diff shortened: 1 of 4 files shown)


program/src/processor.rs
```diff
@@ -29,12 +29,16 @@ use {
...
     if *vote_account_info.owner != solana_vote_program::id() {
         return Err(ProgramError::IncorrectProgramId);
     }
 
+    // JC: I think the code was added after this commit, but can we switch to
+    // `deserialize_into_uninit` to avoid the overhead of creating a
+    // `VoteState::default`?
     let mut vote_state = Box::<VoteState>::default();
     VoteState::deserialize_into(&vote_account_info.data.borrow(), &mut vote_state)
         .map_err(|_| ProgramError::InvalidAccountData)?;
```
2501babe, line 44: yup that is new!

(im going to go through and put comments on anything that requires it, then write my own pr and close this one and we can un-backwards the code review process. assume anything i dont comment on, im agreeing with)

program/src/processor.rs
```diff
@@ -86,8 +100,18 @@ fn relocate_lamports(
...
 // on downstream errors if a required signer is missing the checked mode expects
 // every AccountInfo passed in should be a signer and errors if any is not
+// JC: I kept having to re-read what this was doing to be sure it was ok in
+// every context that it's used.
+// I appreciate that this is more work for you, but could you add a comment for
+// every single time it's called with `checked` as `false` and explain where the
+// signer will be checked? Ie. "unchecked because meta.authorize() will check
+// that the stake or withdraw authority signed".
+// It would certainly make me feel better and help ensure that we preserve all
+// signer checks.
```
2501babe, line 110: this function confuses me too, ill think about a better interface, even if its just replacing the bool with a checked `collect_signers` and a `collect_signers_unchecked`... yea actually im just going to rethink this. i want to check signers when we make the hashmap, instead of worrying about checking them explicitly in every processor and potentially missing one, so its correct by construction. and i did it right but the interface i chose is not good

the problem with `collect_signers` is its serving two masters. in the existing stake program, a _lot_ of process functions dont check signers and rely on `Authorized` to do the checks internally. for those, i check the signers myself, and added `NOTE` comments where i did it differently in case i later realize through testing that i broke compat by being too strict

however `collect_signers` is _also_ used for three unchecked processors (set lockup, authorize, authorize with seed) that have checked variants, where the checked ones require that new authorities sign as a safety feature, while the unchecked ones just take a pubkey. but because of the nature of the `Authorized` functions these two kinds of things get mixed together in the same hashmap. its also confusing because they all take different arguments structs, i even had to go back and look at `sdk/program/src/stake/instruction.rs` to remind myself how these work

ok yea i think i tried to make this too modular, im just going to have two different functions that behave differently to make it make more sense

program/src/processor.rs
```diff
@@ -231,6 +260,8 @@ fn move_stake_or_lamports_shared_checks(
     }
 
     // source and destination must be writable
+    // JC nit: we can probably remove these too since the runtime will definitely
+    // catch this
     if !source_stake_account_info.is_writable || !destination_stake_account_info.is_writable {
         return Err(ProgramError::InvalidInstructionData);
     }
```
2501babe, line 267: this is because move stake and move lamports are defined by simd to abort if theyre not writable and i didnt want to worry about there being any possible codepath that resulted in no attempt to write and a successful no-op. but if we are going to have one impl for all validator clients it matters less

program/src/processor.rs
```diff
@@ -572,6 +610,10 @@ impl Processor {
         // this is somewhat subtle, but if the stake account is Uninitialized, you pass
         // it twice and sign ie, Initialized or Stake, we use real withdraw
         // authority. Uninitialized, stake account is its own authority
+        // JC: do you actually need to pass it twice? I imagine in the current
+        // version, you could just pass it once as a signer and it would Just Work,
+        // but this falls in the same camp as the other similar comments: if
+        // people are depending on that current weird behavior, we break 'em.
```
2501babe, line 616: youre right, i was treating the doc comment on the withdraw instruction literally rather than seriously. they get the account by index off the invoke context in the existing stake program, which i assume means its deduped to the highest allowed privilege

in fact the nature of the transaction format means that its impossible to restrict privlege on the instruction level, doesnt it? the transaction defines what accounts are writable or signers, and each instruction account only had an index into the accounts list. including the same account multiple times with different privileges actually escalates them all to the highest. we dont need to accept the account multiple times at all

program/src/processor.rs
```diff
@@ -763,6 +813,15 @@ impl Processor {
...
 
+        // JC: I hope I'm missing something, but is this instruction currently broken
+        // if there's a lockup in force and you want to change the withdraw
+        // authority? It looks like you're fixing it here!
+        // In the old implementation, I don't think it'll ever add the lockup
+        // authority to the signers hashset.
+        // But to make it consistent with `authorize_checked_with_seed`, how about
+        // passing `true` for checked? I don't think it ends up making a difference,
+        // since if the lockup is not in force anymore, you can just not pass
+        // the custodian at all, and it'll work.
```
2501babe, line 824: i believe youre right... it looks like the existing stake program inits the signers object with all signers. authorize uses this object, but authorize with need ignores it and constructs a new one, and only adds the seed account, so it will fail if theres a lockup. i dont recall noticing it was wrong, i think i just did the right thing automatically once essential validation code was no longer split between two files

program/src/processor.rs
```diff
@@ -819,6 +882,19 @@ impl Processor {
...
+        // In the old version, it might be possible to pass in a self-owned
+        // stake account as a signer, followed by the clock, followed by *anything*,
+        // instead of the old authority as a signer.
+        // Even though it's different, I think we can reasonably go forward with this,
+        // then respond to any complaints with: "No, you're doing it wrong, fix
+        // your code". But we should also warn people. What do you think?
+        // On the other hand, we could also pass all of the account infos here?
+        // Note that it might also be possible to pass in the old authority *after*
+        // the lockup authority as a signer and also work, so we might be breaking
+        // some weird cases even if we do that.
```
2501babe, line 897: hm maybe we do have to do it the old way then... i dont like that it essentially means we dont have a well-defined interface, but if we dont i guess we dont

program/src/processor.rs
joncinque: Ah ok, if it's in the SIMD, let's leave as is

program/src/processor.rs
joncinque: Yep that's exactly correct

program/src/processor.rs
joncinque: This might be the opportunity to strictly define the interface. We'll just need to be aware that it might be a source of divergence while fuzzing the bpf program to the native version.

2501babe:
continuing this in #11. i fixed all the trivial stuff in the first commit and also left comments on anything you commented on that i havent fixed yet. so im going to close this

PR #11, opened 2024-10-23
program: respond to jon code review

2501babe:
address points raised in initial code review:
* convert `InstructionError` to `ProgramError` without a trait
* use `VoteState::deserialize_into_uninit()`
* drop one lamport borrow before taking another
* rewrite how signers are handled to be clearer and also preserve unfortunate native stake program permissiveness
* add comments for skipped owner checks

also some of the block comments changed but thats because there was a rustfmt rule that messed with comments, which i didnt notice garbled them in a push to master commit lol

after this pr, the most important thing for me to do with the stake program is get solfuzz-agave up and running. currently this repo has better (exhaustive tests for all stake account state combinations for complex operations like split and merge, exhaustive tests for missing signers) but also worse (many of the ad hoc tests for edge cases have not been ported, and lockup coverage is very lacking) test coverage than the monorepo stake program. i believe tho that firedancer ported all the tests in `stake_instruction.rs` to test fixtures. so if we pass those, i will be very confident my implementation is correct

also i think i will implement the sysvar skipping logic, theres no downside to it and after we launch we can remove all those accounts from the instruction builders. but i want to do this after solfuzz-agave because it might cause false negatives if there are tests that expect omitting a sysvar to cause a failed transaction

### DIFF
--- a/program/src/processor.rs
+++ b/program/src/processor.rs
@@ -1,5 +1,5 @@
 use {
-    crate::{helpers::*, id, PERPETUAL_NEW_WARMUP},
+    crate::{helpers::*, id, PERPETUAL_NEW_WARMUP_COOLDOWN_RATE_EPOCH},
     solana_program::{
         account_info::{next_account_info, AccountInfo},
         clock::Clock,
@@ -22,7 +22,7 @@ use {
         sysvar::{epoch_rewards::EpochRewards, stake_history::StakeHistorySysvar, Sysvar},
         vote::{program as solana_vote_program, state::VoteState},
     },
-    std::collections::HashSet,
+    std::{collections::HashSet, mem::MaybeUninit},
 };
 
 // TODO a nice change would be to pop an account off the queue and discard if
@@ -30,14 +30,18 @@ use {
 // without breaking compat to be done after release, we keep the existing
 // interface for all instructions for compat with firedancer
 
+// TODO undecided if we do it in initial release or in a future release...
+// but we could use StakeError much more extensively and impl PrintProgramError
+
 fn get_vote_state(vote_account_info: &AccountInfo) -> Result<Box<VoteState>, ProgramError> {
     if *vote_account_info.owner != solana_vote_program::id() {
         return Err(ProgramError::IncorrectProgramId);
     }
 
-    let mut vote_state = Box::<VoteState>::default();
-    VoteState::deserialize_into(&vote_account_info.data.borrow(), &mut vote_state)
+    let mut vote_state = Box::new(MaybeUninit::uninit());
+    VoteState::deserialize_into_uninit(&vote_account_info.data.borrow(), vote_state.as_mut())
         .map_err(|_| ProgramError::InvalidAccountData)?;
+    let vote_state = unsafe { Box::from_raw(Box::into_raw(vote_state) as *mut VoteState) };
 
     Ok(vote_state)
 }
@@ -69,15 +73,19 @@ fn relocate_lamports(
     destination_account_info: &AccountInfo,
     lamports: u64,
 ) -> ProgramResult {
-    let mut source_lamports = source_account_info.try_borrow_mut_lamports()?;
-    **source_lamports = source_lamports
-        .checked_sub(lamports)
-        .ok_or(ProgramError::InsufficientFunds)?;
+    {
+        let mut source_lamports = source_account_info.try_borrow_mut_lamports()?;
+        **source_lamports = source_lamports
+            .checked_sub(lamports)
+            .ok_or(ProgramError::InsufficientFunds)?;
+    }
 
-    let mut destination_lamports = destination_account_info.try_borrow_mut_lamports()?;
-    **destination_lamports = destination_lamports
-        .checked_add(lamports)
-        .ok_or(ProgramError::ArithmeticOverflow)?;
+    {
+        let mut destination_lamports = destination_account_info.try_borrow_mut_lamports()?;
+        **destination_lamports = destination_lamports
+            .checked_add(lamports)
+            .ok_or(ProgramError::ArithmeticOverflow)?;
+    }
 
     Ok(())
 }
@@ -86,6 +94,31 @@ fn relocate_lamports(
 // constructs it the unchecked mode doesnt add pubkeys of non-signers, relying
 // on downstream errors if a required signer is missing the checked mode expects
 // every AccountInfo passed in should be a signer and errors if any is not
+//
+// FIXME do something with this, but after easy changes, since i need to change
+// all callers i should have two functions, since this is used for:
+// * check all signers. easy
+// * check *some* both not *all* signers, for Initializ/Authorize/SetLockup
+//   whose checked variants check signer for accounts they dont actually need
+//   just as a safety feature
+// also rename optional_account to be more descriptive
+//
+// XXX ok so functions that create their own signers in stake_state.rs:
+// * auth seed, but its broken
+// * withdraw, only adds withdraw authority
+//   but then it gets the lockup authority and checks without hashset
+// * move stake/lamps, only add staker authority
+// everything else gets it from stake_instruction.rs
+// i need to go through every instruction and double check account len asserts
+// there are some that may "work" with fewer than expected accounts
+// abusing the fact that index lookups can get them out of order
+//
+// XXX but yes i need to contemplate my signer collection interface
+// sigh... every other instruction just hoovers up all the signers
+// so undisciplined. well the good news is at least i can be lazier with the code
+// just loop accounts twice. once get the signers once pop the accounts
+// i think we should break people who omit accounts relying on the buggy interface
+// need to record which if any instructions allow that tho
 fn collect_signers<'a>(
     accounts: &[&'a AccountInfo],
     optional_account: Option<&'a AccountInfo>,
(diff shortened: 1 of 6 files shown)


program/src/processor.rs
```diff
@@ -560,34 +575,34 @@ impl Processor {
...
-            option_lockup_authority_info,
-            true,
-        )?;
+        let clock = &Clock::from_account_info(clock_info)?;
+        let stake_history = &StakeHistorySysvar(clock.epoch);
+
+        // this is somewhat subtle. for Initialized and Stake, there is a real authority
+        // but for Uninitialized, the source account is passed twice, and signed for
+        let (signers, custodian) =
+            collect_signers_checked(Some(withdraw_authority_info), option_lockup_authority_info)?;
```
2501babe, line 592: reproducing your comment from #7:
```
        // JC: do you actually need to pass it twice? I imagine in the current
        // version, you could just pass it once as a signer and it would Just Work,
        // but this falls in the same camp as the other similar comments: if
        // people are depending on that current weird behavior, we break 'em.
```

`Withdraw` has the dubious distinction of being the only stake program instruction (other than the relatively new `Move` instructions) that builds its own signers hashset and does it correctly. the `AuthorizeWithSeed` instructions do it wrong, and everything else uses the globbed hashset of all signers

you actually do need to pass the account twice for the `Uninitialized` case because the native program asserts there are at least five accounts and checks signer on the fifth. if you pass the same account twice with different `AccountMeta` privileges, it elevates the account to the highest, and works fine. if you pass the account once, it fails for not having enough accounts. if you pass a different account the second time, it fails to find the stake account address in the signers hashset and fails

2501babe: also i left the `if !signers.contains(source_stake_account_info.key) { ... }` block unchanged because this is already confusing enough

program/src/processor.rs
```diff
@@ -22,22 +21,26 @@ use {
...
     if *vote_account_info.owner != solana_vote_program::id() {
         return Err(ProgramError::IncorrectProgramId);
     }
 
-    let mut vote_state = Box::<VoteState>::default();
-    VoteState::deserialize_into(&vote_account_info.data.borrow(), &mut vote_state)
+    let mut vote_state = Box::new(MaybeUninit::uninit());
+    VoteState::deserialize_into_uninit(&vote_account_info.try_borrow_data()?, vote_state.as_mut())
         .map_err(|_| ProgramError::InvalidAccountData)?;
+    let vote_state = unsafe { Box::from_raw(Box::into_raw(vote_state) as *mut VoteState) };
```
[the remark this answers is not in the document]
joncinque: Considering your code is essentially the exact implementation of `Box<MaybeUninit<T>>::assume_init()` at https://doc.rust-lang.org/src/alloc/boxed.rs.html#937, this looks correct to me!

More seriously, I had to do the same thing of replicating a function that wasn't available yet on MaybeUninit for slices, so this approach is valid.

The one thing I find weird is calling `as_mut()` to convert from `Box<MaybeUninit<T>>` to `&mut MaybeUninit<T>` when calling `deserialize_into_uninit`, only for that function to call `as_mut_ptr()` for the private `deserialize_into_ptr` function, when it would be a bit more ergonomic to call directly into `deserialize_into_ptr`, but that's not a problem with this code. And I imagine that the compiler just optimizes that away.

joncinque (commented):
Looks really good! Just a question on some of the uses of `next_account_info(...).ok()`

program/src/processor.rs
```diff
@@ -643,29 +658,28 @@ impl Processor {
...
 
     fn process_deactivate(accounts: &[AccountInfo]) -> ProgramResult {
+        let signers = collect_signers(accounts);
         let account_info_iter = &mut accounts.iter();
+
         let stake_account_info = next_account_info(account_info_iter)?;
         let clock_info = next_account_info(account_info_iter)?;
-        let clock = &Clock::from_account_info(clock_info)?;
-        let stake_authority_info = next_account_info(account_info_iter)?;
+        let _option_stake_authority_info = next_account_info(account_info_iter).ok();
```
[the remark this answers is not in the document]
2501babe: actually maybe it would make more sense if i removed the `.ok()`. this and other similar lines are a result of, i did my best to thoroughly check exactly how many accounts the native stake program asserts exist, and the purpose of the line is just to say "this account may be here, in fact is probably here, although we dont need it directly"

99% of cases it should exist, be a signer, and we collect its pubkey in the `let signers = ...` line. for self-signed stake accounts it may not exist, and since the native program doesnt assert it exists, we cannot unwrap or we break the interface

youre right deleting the line would have the same effect but i think having it (without `.ok()` is better to say "this is what the interface is supposed to be." i think my game plan long term is after we launch a native-compat bpf stake, ill just set up a validator with a geyser plugin to find out whether anyone _actually_ does the things they technically can do, and break anything that is rare or nonexistant in a future update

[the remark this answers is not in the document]
2501babe: after thinking about this for 30 seconds more i guess a comment would suffice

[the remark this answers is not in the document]
2501babe: actually there are a couple more of these where the native program fails to assert accounts exist but i assert they do. what im going to do is go through all these again, annotate every function with a comment describing what the native program actually does, and put commented-out lines at the end indicating what extra accounts we would like to be there. we do want comments, rather than actually calling the functions, because even if we dont use them they mutate the accounts iter and so cannot be optimized away by rustc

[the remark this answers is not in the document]
2501babe: also just noting this is and issue in the realm of "well, lets be as permissive as the native program just in case" and not "oh no we overlooked a major flaw." i honestly dont think many if any of these "weird" transactions exist, since they would require someone to implement their own transaction builders, for no real purpose

2501babe:
ok! after going through every instruction one by one, i believe signers handling now matches the native program exactly (aside from fixing authorize with seed)

program/src/processor.rs
```diff
@@ -325,28 +369,33 @@ impl Processor {
...
+        // however, the fifth account used to be stake config, and it was removed without changing the assert
+        // there may be existing workflows that place stake authority fifth, or sixth, or omit it and self-sign
+        // we preserve the fifth unwrap above to avoid making this processor even more permissive
+        // let _stake_authority_info = next_account_info(account_info_iter);
 
-        // NOTE the existing program behaves as if this were false
-        // it should not break compat to check here, but may change errors
-        let (signers, _) = collect_signers(&[stake_authority_info], None, true)?;
+        let clock = &Clock::from_account_info(clock_info)?;
+        let stake_history = &StakeHistorySysvar(clock.epoch);
```
2501babe, line 390: im going to do a pr after this one to implement a version of `next_account_info()` which skips sysvars and stake config. this should be a perf improvement for 100% of transactions, once people omit the accounts, because getting sysvars by sycall is cheaper than building the `AccountInfo` in `entrypoint!`. once this program is live we can change the instruction builders to not include sysvars and backport all the way to 2.0 no matter what version is on mainnet

once i make that change, this code becomes:
* return error if fewer than five accounts
* collect signers
* create accounts iter
* unwrap four accounts

program/src/processor.rs
```diff
@@ -325,28 +369,29 @@ impl Processor {
...
         let vote_account_info = next_account_info(account_info_iter)?;
         let clock_info = next_account_info(account_info_iter)?;
-        let clock = &Clock::from_account_info(clock_info)?;
         let _stake_history_info = next_account_info(account_info_iter)?;
         let _stake_config_info = next_account_info(account_info_iter)?;
-        let stake_authority_info = next_account_info(account_info_iter)?;
 
-        let stake_history = &StakeHistorySysvar(clock.epoch);
+        // other accounts
+        // let _stake_authority_info = next_account_info(account_info_iter);
```
2501babe, line 383: im going to do a pr after this one to implement a version of `next_account_info()` which skips sysvars and stake config. this should be a perf improvement for 100% of transactions, once people omit the accounts, because getting sysvars by syscall is cheaper than building the `AccountInfo` in `entrypoint!`. once this program is live we should change the instruction builders to omit sysvars and backport all the way to 2.0 no matter what version is on mainnet

once i replace `next_account_info()`, this code becomes:
* return error if fewer than three accounts
* collect signers
* create accounts iter
* unwrap two accounts

then if we go forward with breaking interface changes, we would unwrap a third account, the stake authority, and it will do the right thing regardless of whether stake authority is third or sixth

i point this out here (rather than for all the other instructions where authority is technically optional) because the monorepo instruction builder for `Delegate` still includes stake config for "backwards compatibility," even tho its not technically necessary because stake authority is not asserted to exist, but we should leave it as-is to avoid making this even more complicated

in other words no one using our instruction builders would ever place stake authority fifth

joncinque: Just be careful -- skipping the sysvar accounts is a breaking change between this and native, since at the very least we should check the sysvar account address.

Unless there's a security reason, we should not break the instruction interface, and instead introduce new instructions, but have the old instruction creators create the new instructions -- ie `fn delegate()` creates a `DelegateNew` instruction.

There's a lot of staking tooling / programs that rely on the current instructions, so we can't break them. After we introduce the new instructions, we can give people a couple of years before removing them.

joncinque (approved):
These changes look good to me!

program/src/processor.rs
2501babe: what i mean is write a `next_account_info()` that does something like this:
```rust
fn hana_next_account(account_iter: &mut _) -> ProgramResult<AccountInfo> {
  loop {
    let account = next_account_info(account_iter);
    match &account {
      Ok(a) if is_sysvar_or_stake_config(a) => continue,
      _ => return account,
    }
  }
}
```

in other words, we dont change the positionality of any accounts, we just advance past ones we dont care about. that way we can support old and new instructions in a non-breaking way (unless you are thinking of a different kind of breaking change)

im also not opposed to just making a second brand new interface that does everything right and deprecating the old one tho

joncinque: Ah ok, that makes sense! Sorry, I was worried that you were suggesting breaking changes on the old instructions. 

Both approaches are fine by me, we can see what it looks like when we get there. I did really like how you implemented move-stake / move-lamports, so a brand new interface might be the better option.

PR #13, opened 2024-11-01
Bump dependencies

febo:
### Problem

Local build/CI are currently not working since `v2.1` moved to a branch now.

### Solution

Bump the dependencies for both the program and Rust client to use the version from `crates.io`.

The generated Rust client changed since the `@codama/renderer-rust` version was bumped to remove the dependency on Anchor.

### DIFF
--- a/clients/rust/Cargo.toml
+++ b/clients/rust/Cargo.toml
@@ -17,10 +17,10 @@ num-derive = "^0.3"
 num-traits = "^0.2"
 serde = { version = "^1.0", features = ["derive"], optional = true }
 serde_with = { version = "^3.0", optional = true }
-solana-program = { git = "https://github.com/anza-xyz/agave.git", version = "^2.1", branch = "v2.1" }
+solana-program = ">= 1.16, < 3"
 thiserror = "^1.0"
 
 [dev-dependencies]
 assert_matches = "1.5.0"
-solana-program-test = { git = "https://github.com/anza-xyz/agave.git", version = "^2.1", branch = "v2.1" }
-solana-sdk = { git = "https://github.com/anza-xyz/agave.git", version = "^2.1", branch = "v2.1" }
+solana-program-test = "2.1"
+solana-sdk = "2.1"
--- a/clients/rust/Cargo.toml
+++ b/clients/rust/Cargo.toml
@@ -17,7 +17,7 @@ num-derive = "^0.3"
 num-traits = "^0.2"
 serde = { version = "^1.0", features = ["derive"], optional = true }
 serde_with = { version = "^3.0", optional = true }
-solana-program = ">= 1.16, < 3"
+solana-program = "2.1"
 thiserror = "^1.0"
 
 [dev-dependencies]
--- a/clients/rust/Cargo.toml
+++ b/clients/rust/Cargo.toml
@@ -8,27 +8,19 @@ readme = "README.md"
 license-file = "../../LICENSE"
 
 [features]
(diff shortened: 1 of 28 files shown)


joncinque (commented):
Rather than using the git dependencies, can we just use the ones from crates.io?

febo:
> Rather than using the git dependencies, can we just use the ones from crates.io?

Good point! Do we need to pin them to `=2.1`?

joncinque:
We shouldn't need to, since we *should* adhere to semver from now on

2501babe:
yep the goal was to go to 2.1 (no `=` required) once it was on crates, we couldnt use 2.0 because it doesnt have the interface for getting stake history via syscall

2501babe:
should also mean if/when there is an anchor on 2.1 we can bring the rust client back into the workspace and start using ci for real

febo:
> We shouldn't need to, since we _should_ adhere to semver from now on

Unfortunately we can't use individual crate dependencies on the Rust client yet, since the generated code "assumes" that you have `solana-program`. I created an issues in Codama to track this (https://github.com/codama-idl/codama/issues/277).

febo:
> yep the goal was to go to 2.1 (no `=` required) once it was on crates, we couldnt use 2.0 because it doesnt have the interface for getting stake history via syscall

We could replace the `solana-program` dependency for individual crates. I did not include this is this PR since it will involve changes the program source.

clients/rust/Cargo.toml
```diff
@@ -8,27 +8,19 @@ readme = "README.md"
...
 
 [dependencies]
-anchor-lang = { version = "0.30.0", optional = true }
 borsh = "^0.10"
 num-derive = "^0.3"
 num-traits = "^0.2"
 serde = { version = "^1.0", features = ["derive"], optional = true }
 serde_with = { version = "^3.0", optional = true }
-solana-program = { git = "https://github.com/anza-xyz/agave.git", version = "=2.1.0" }
+solana-program = ">= 1.16, < 3"
```
joncinque, line 20: This range might be a bit big, is there a reason for it?

program/Cargo.toml
```diff
@@ -18,21 +18,15 @@ borsh = { version = "1.5.1", features = ["derive", "unstable__schema"] }
...
-solana-program = { git = "https://github.com/anza-xyz/agave.git", version = "=2.1.0" }
+solana-program = "2.1"
 thiserror = "1.0.63"
 
 [dev-dependencies]
-solana-program-test = { git = "https://github.com/anza-xyz/agave.git", version = "=2.1.0" }
-solana-sdk = { git = "https://github.com/anza-xyz/agave.git", version = "=2.1.0" }
-solana-vote-program = { git = "https://github.com/anza-xyz/agave.git", version = "=2.1.0" }
+solana-program-test = "2.1"
+solana-sdk = "=2.1"
```
joncinque, line 26: This probably shouldn't pin just one crate:
```suggestion
solana-sdk = "2.1"
```

joncinque (commented):
Looks good! Just one little thing

program/Cargo.toml
febo: Yep, that was a typo.

clients/rust/Cargo.toml
febo: No particular reason, do you have a suggestion of which range might be more appropriate?

agca52 (approved):
Thankk you

PR #12, opened 2024-11-01
interface: Add stake files from sdk with history

febo:
#### Problem

The initial PR #10 adds the Stake interface files from the SDK without their git history.

#### Summary of changes

Copy the files from the SDK together with their associated history using `git-filter-repo`:

```
# add the remote for the source repo
git remote add agave https://github.com/anza-xyz/agave.git
# fetch the master from agave under a different local branch
git fetch agave master:stake-from-agave
git switch stake-from-agave
# be safe, remove the remote
git remote rm agave
# filter the stake related files
python3 ../git-filter-repo \
--path sdk/program/src/stake/config.rs \
--path sdk/program/src/stake/instruction.rs  \
--path sdk/program/src/stake/mod.rs \
--path sdk/program/src/stake/stake_flags.rs \
--path sdk/program/src/stake/state.rs \
--path sdk/program/src/stake/tools.rs \
--path sdk/program/src/stake_history.rs \
--path-rename sdk/program/src/stake/config.rs:interface/src/config.rs \
--path-rename sdk/program/src/stake/instruction.rs:interface/src/instruction.rs \
--path-rename sdk/program/src/stake/mod.rs:interface/src/lib.rs \
--path-rename sdk/program/src/stake/stake_flags.rs:interface/src/stake_flags.rs \
--path-rename sdk/program/src/stake/state.rs:interface/src/state.rs \
--path-rename sdk/program/src/stake/tools.rs:interface/src/tools.rs \
--path-rename sdk/program/src/stake_history.rs:interface/src/stake_history.rs \
--force
# re-add this repo
git remote add origin https://github.com/febo/stake.git
git fetch origin master
git rebase master
```

PR #10, opened 2024-10-18
Add stake interface

febo:
This PR adds the `interface` crate for the Stake program, moving the Stake types and instruction from the monorepo.

PR #16, opened 2024-11-12
interface: Re-add stake files from sdk with history

febo:
### Problem

PR #12 was "squashed & merged", which does not preserve the history.

### Solution

This PR re-adds the files so they can be "rebased & merged" instead. PR #15 reverted #12.


joncinque (approved):
Looks good! For reference, the PR that was landed and reverted is #12, not #10

PR #17, opened 2024-11-15
Add interface crate

febo:
### Problem

While the repository contains the source files for the Stake interface, it does not define the interface crate.

### Solution

This PR adds the interface `Cargo.toml` configuration and updates its dependencies to use the individual crates.

There are 2 aspects that will be handled in separate PRs:

1. Interface scripts for the CI: this will be added once https://github.com/solana-program/system/pull/23 is merged, since it will follow a similar pattern.
2. Move the [`StakeHistory`](https://github.com/anza-xyz/agave/blob/master/sdk/program/src/sysvar/stake_history.rs) sysvar into this repository.

### DIFF
--- a/interface/Cargo.toml
+++ b/interface/Cargo.toml
@@ -11,7 +11,7 @@ license-file = "../LICENSE"
 borsh = { version = "1.5.1", features = ["derive", "unstable__schema"], optional = true }
 borsh0-10 = { package = "borsh", version = "0.10.3", optional = true }
 num-traits = "0.2"
-serde = { version = "1.0.210", features = ["derive"], optional = true }
+serde = { version = "1.0.210", optional = true }
 serde_derive = { version = "1.0.210", optional = true }
 solana-decode-error = "^2.1"
 solana-clock = "^2.1"
@@ -21,7 +21,7 @@ solana-frozen-abi-macro = { version = "^2.1", features = ["frozen-abi"], optiona
 solana-instruction = "^2.1"
 solana-program-error = { version = "^2.1", optional = true }
 solana-pubkey = { version = "^2.1", default-features = false }
-solana-system-interface = { version = "0.0.1", git = "https://github.com/solana-program/system.git" }
+solana-system-interface = "^1.0"
 
 [dev-dependencies]
 assert_matches = "1.5.0"
@@ -34,6 +34,8 @@ strum_macros = "0.24"
 
 [package.metadata.docs.rs]
 targets = ["x86_64-unknown-linux-gnu"]
+all-features = true
+rustdoc-args = ["--cfg=docsrs"]
 
 [features]
 bincode = [
--- a/interface/Cargo.toml
+++ b/interface/Cargo.toml
@@ -1,0 +1,51 @@
+[package]
+name = "solana-stake-interface"
+version = "0.1.0"
+description = "Instructions and constructors for the Stake program"
+repository = "https://github.com/solana-program/stake"
+edition = "2021"
+readme = "README.md"
+license-file = "../LICENSE"
+
+[dependencies]
+borsh = { version = "1.5.1", features = ["derive", "unstable__schema"], optional=true }
+borsh0-10 = { package = "borsh", version = "0.10.3", optional=true }
+num-traits = "0.2"
+serde = { version = "1.0.210", features = ["derive"] }
+serde_derive = "1.0.210"
+solana-decode-error = "^2.1"
+solana-clock = "^2.1"
+solana-cpi = "^2.1"
+solana-frozen-abi = { version = "^2.1", features = ["frozen-abi"], optional = true }
+solana-frozen-abi-macro = { version = "^2.1", features = ["frozen-abi"], optional = true }
+solana-instruction = { version = "^2.1", features = ["bincode", "std"] }
+solana-program-error = "^2.1"
+solana-pubkey = { version = "^2.1", default-features = false, features = ["serde"] }
+solana-system-interface = { version = "0.0.1", git = "https://github.com/solana-program/system.git" }
+
+[dev-dependencies]
+assert_matches = "1.5.0"
+bincode = "1.3.3"
+solana-borsh = "^2.1"
+solana-program = { version = "^2.1", default-features = false }
+static_assertions = "1.1.0"
+strum = "0.24"
+strum_macros = "0.24"
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
+
+[features]
+borsh = [
+    "dep:borsh",
+    "dep:borsh0-10",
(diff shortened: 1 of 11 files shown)


interface/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+edition = "2021"
+readme = "README.md"
+license-file = "../LICENSE"
+
+[dependencies]
+borsh = { version = "1.5.1", features = ["derive", "unstable__schema"], optional=true }
+borsh0-10 = { package = "borsh", version = "0.10.3", optional=true }
+num-traits = "0.2"
+serde = { version = "1.0.210", features = ["derive"] }
+serde_derive = "1.0.210"
```
joncinque, line 15: Is it possible to make these optional with a `serde` crate feature, as we've been doing? And then we can enable the `bincode` / `serde` features on dependencies optionally. It'll be a little messy becauase all of the instruction creators require bincode / serde, but this way, in the future we can move the instruction creators to use something else.

interface/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+solana-decode-error = "^2.1"
+solana-clock = "^2.1"
+solana-cpi = "^2.1"
+solana-frozen-abi = { version = "^2.1", features = ["frozen-abi"], optional = true }
+solana-frozen-abi-macro = { version = "^2.1", features = ["frozen-abi"], optional = true }
+solana-instruction = { version = "^2.1", features = ["bincode", "std"] }
+solana-program-error = "^2.1"
+solana-pubkey = { version = "^2.1", default-features = false, features = ["serde"] }
+solana-system-interface = { version = "0.0.1", git = "https://github.com/solana-program/system.git" }
```
joncinque, line 24: side note: is this ready to be published so that we can use the version from crates.io instead?

interface/src/error.rs
```diff
@@ -0,0 +1,242 @@
...
+            11 => StakeError::MinimumDelinquentEpochsForDeactivationNotMet,
+            12 => StakeError::InsufficientDelegation,
+            13 => StakeError::RedelegateTransientOrInactiveStake,
+            14 => StakeError::RedelegateToSameVoteAccount,
+            15 => StakeError::RedelegatedStakeMustFullyActivateBeforeDeactivationIsPermitted,
+            16 => StakeError::EpochRewardsActive,
+            _ => panic!("unsupported error code"),
+        }
+    }
+}
```
joncinque, line 225: Is this implementation needed? I don't see where we had it before. With `FromPrimitive` and `ToPrimitive`, we should be covered, and the `panic!` here doesn't inspire confidence :sweat_smile:

interface/src/instruction.rs
```diff
@@ -924,7 +850,10 @@ pub fn move_lamports(
 
 #[cfg(test)]
 mod tests {
-    use {super::*, crate::instruction::InstructionError};
+    use {
+        crate::error::StakeError, solana_decode_error::DecodeError,
+        solana_instruction::error::InstructionError,
+    };
```
joncinque, line 856: nit: since these tests don't use the instructions at all, how about moving them into `error.rs`?

interface/src/instruction.rs
```diff
@@ -287,7 +206,7 @@ pub enum StakeInstruction {
     ///   0. `[WRITE]` Delegated stake account
     ///   1. `[]` Delinquent vote account for the delegated stake account
     ///   2. `[]` Reference vote account that has voted at least once in the last
-    ///      `MINIMUM_DELINQUENT_EPOCHS_FOR_DEACTIVATION` epochs
+    ///      [`super::MINIMUM_DELINQUENT_EPOCHS_FOR_DEACTIVATION`] epochs
```
joncinque, line 209: nit: how about `crate::` rather than `super::`?

joncinque (commented):
Looks great overall! Mostly small things

interface/src/instruction.rs
febo: Good idea.

interface/Cargo.toml
febo: How would that work? Do we make the whole instruction creator optional as well under a different feature (e.g., `"bincode"`)? So enabling `"bincode"` would enable `"serde"`, since `Instruction::new_with_bincode` needs both.

joncinque: Yes exactly

interface/Cargo.toml
```diff
@@ -0,0 +1,59 @@
...
+edition = "2021"
+readme = "README.md"
+license-file = "../LICENSE"
+
+[dependencies]
+borsh = { version = "1.5.1", features = ["derive", "unstable__schema"], optional = true }
+borsh0-10 = { package = "borsh", version = "0.10.3", optional = true }
+num-traits = "0.2"
+serde = { version = "1.0.210", features = ["derive"], optional = true }
+serde_derive = { version = "1.0.210", optional = true }
```
joncinque, line 15: You don't need the `derive` feature if you're bringing in `serde_derive`
```suggestion
serde = { version = "1.0.210", optional = true }
serde_derive = { version = "1.0.210", optional = true }
```

interface/Cargo.toml
```diff
@@ -0,0 +1,59 @@
...
+assert_matches = "1.5.0"
+bincode = "1.3.3"
+solana-borsh = "^2.1"
+solana-program = { version = "^2.1", default-features = false }
+static_assertions = "1.1.0"
+strum = "0.24"
+strum_macros = "0.24"
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
```
joncinque, line 36: Wanna add the extra docs bits here?
```suggestion
[package.metadata.docs.rs]
targets = ["x86_64-unknown-linux-gnu"]
all-features = true
rustdoc-args = ["--cfg=docsrs"]
```

interface/src/stake_history.rs
```diff
@@ -1,19 +1,19 @@
 //! A type to hold data for the [`StakeHistory` sysvar][sv].
 //!
 //! [sv]: https://docs.solanalabs.com/runtime/sysvars#stakehistory
-//!
-//! The sysvar ID is declared in [`sysvar::stake_history`].
-//!
-//! [`sysvar::stake_history`]: crate::sysvar::stake_history
```
joncinque, line 7: You can do this in follow-up work, but how about bringing in `solana-sysvar-id` crate and implementing it for stake history in this crate?

joncinque (commented):
Looks really close! Just a few last bits that I noticed on this pass

interface/src/stake_history.rs
febo: Yeah, I will bring the stake history sysvar on a separate PR. 👍

joncinque (approved):
Beautiful!

PR #18, opened 2024-12-02
clients: Add client generation

febo:
### Problem

Currently, the stake program does not include annotation to generate an IDL. This prevents using Codama to generate clients.

### Solution

Use `Shank` to annotate the program and generate clients.

**Note:** Majority of source files under the `clients` folder are auto-generated.


clients/js/package.json
```diff
@@ -34,13 +34,13 @@
   },
   "license": "MIT",
   "peerDependencies": {
-    "@solana/web3.js": "2.0.0-rc.1"
+    "@solana/web3.js": "2.0.0"
```
steveluscher, line 37: Is this how we do this? Not `^2`?

steveluscher (approved):
ezgif-1-a6c9ea68b9

clients/js/package.json
febo: You are right, there is a missing `^` - fixed!

interface/src/instruction.rs
```diff
@@ -56,6 +69,30 @@ pub enum StakeInstruction {
...
+    #[account(
+        0,
+        writable,
+        name = "stake",
+        desc = "Stake account to be updated"
+    )]
+    #[account(
+        1,
+        name = "clock_sysvar",
+        desc = "Rent sysvar"
```
joncinque, line 81:
```suggestion
        desc = "Clock sysvar"
```

joncinque (commented):
Just the one nit.

It doesn't need to be done in this PR, but what do you think about introducing a feature like `idl` and gating all shank usage on that flag?

That way, users of the crate don't have to pull in shank, and even better, it should make it easier to swap to codama macros in the future with less disruptive breakage.

febo:
> Just the one nit.
> 
> It doesn't need to be done in this PR, but what do you think about introducing a feature like `idl` and gating all shank usage on that flag?
> 
> That way, users of the crate don't have to pull in shank, and even better, it should make it easier to swap to codama macros in the future with less disruptive breakage.

Good point! We could add `shank` as an optional dependency, but theoretically we don't need to add a feature for it. The annotation is only used by `shank` cli, so they don't do anything during compilation.
```
shank = { version = "0.4.2", optional = true }
```
This is enough to not pull in `shank`. The IDL generation still works without the need for a feature. Or we could add a feature, but it won't get used.

joncinque:
I'm not sure I see how it works without a feature (unless you mean using the implicit `dep:shank` feature), but I'll wait to see the code!

febo:
> I'm not sure I see how it works without a feature (unless you mean using the implicit `dep:shank` feature), but I'll wait to see the code!

Hmmm, you are right. The IDL generation works without having a feature, but you will need it to be able to have the validation when you add them to your struct/enum types.

I will add the `idl` feature.

febo:
> > I'm not sure I see how it works without a feature (unless you mean using the implicit `dep:shank` feature), but I'll wait to see the code!
> 
> Hmmm, you are right. The IDL generation works without having a feature, but you will need it to be able to have the validation when you add them to your struct/enum types.
> 
> I will add the `idl` feature.

This did not work. `Shank` CLI does not "detect" the macros when their are behind a feature flag. I removed `Shank` macros and dependency but kept the IDL, so we can have the generated clients.

interface/src/instruction.rs
```diff
@@ -36,6 +36,7 @@ const STAKE_HISTORY_ID: Pubkey =
     derive(serde_derive::Deserialize, serde_derive::Serialize)
 )]
 #[derive(Debug, PartialEq, Eq, Clone)]
+#[rustfmt::skip]
```
joncinque, line 39: nit: this shouldn't be needed anymore

joncinque (approved):
Just the tiny nit, then this can go in!

interface/src/instruction.rs
febo: Good catch!

PR #24, opened 2024-12-06
Stake instruction tests

2501babe:
port _all_ tests (except one that makes no sense) from monorepo stake_instruction.rs. we accomplish this by:
* copying all tests essentially as-is
* changing all uses of `InstructionError` to `ProgramError`
* replacing the internals of `process_instruction`, `process_instruction_as_one_arg`, etc to use mollusk to emulate the previous raw `InvokeContext` creation/usage
* changing those functions to accept a mollusk runner instead of a featureset
* deleting all minimum delegation `test_case` variants. because we no longer have features, minimum delegation is hardcoded as off in the program and cannot be tested. if we ever get consensus on activating it, we can reenable by adding `test_case` using whatever mechanism is chosen
* adding `test_case` variants for bpf stake and native stake. this ensures the same tests pass against both programs
* slightly modifying some tests that pass in `StakeHistory::default()`. the details are somewhat arcane and do not (and could never possibly) occur on a live cluster, but these tests set up an invalid `StakeHistory` full of epoch 0 entries, which the stake history syscall rejects as invalid, when what they really want is a nonexistant stake history

there are two (and only two) places where we behave differently depending on which program is being tested. this is because we fixed a minor bug and expect to see an `InvalidAccountOwner` error instead of `IncorrectProgramId` when an owner is invalid. these can be seen by grepping for `is_bpf()`


program/tests/stake_instruction.rs
```diff
@@ -0,0 +1,7076 @@
...
+    mollusk
+}
+
+fn mollusk_bpf() -> Mollusk {
+    let mut mollusk = Mollusk::new(&id(), "solana_stake_program");
+    mollusk
+        .feature_set
+        .deactivate(&stake_raise_minimum_delegation_to_1_sol::id());
+    mollusk
+}
```
buffalojoec, line 65: Sorry! I need to update the documentation, but you can actually use one Mollusk instance if you wanted. It doesn't map to a program anymore. `new` is just a convenience method to create a new instance and load a BPF program into the cache, but a Mollusk instance comes with a loaded cache and can be pointed at any program in the cache, doesn't matter if you use `default` or `new`.

The program to invoke is dictated by the instruction's program ID.

program/tests/stake_instruction.rs
```diff
@@ -0,0 +1,7076 @@
...
+                } else if rewards::check_id(pubkey) {
+                    create_account_shared_data_for_test(&rewards::Rewards::new(0.0))
+                } else if stake_history::check_id(pubkey) {
+                    create_account_shared_data_for_test(&StakeHistory::default())
+                } else if stake_config::check_id(pubkey) {
+                    config::create_account(0, &stake_config::Config::default())
+                } else if epoch_schedule::check_id(pubkey) {
+                    create_account_shared_data_for_test(&EpochSchedule::default())
+                } else if rent::check_id(pubkey) {
+                    create_account_shared_data_for_test(&Rent::default())
```
buffalojoec, line 167: It's still a bit annoying to do, but you do have this API for keyed sysvar accounts available from the library:
https://github.com/buffalojoec/mollusk/blob/0c7b0edfe5788e9a6892d2c4eb60d43390c7175f/harness/src/sysvar.rs#L101-L134

buffalojoec (commented):
Woohoo!

program/tests/stake_instruction.rs
2501babe: i believe what im doing here is correct (note that this branch uses a local version of mollusk on my machine that includes all three of my open prs). `mollusk_native()` loads no program so it uses the native stake program, `mollusk_bpf()` loads the bpf stake program which overrides the native. they have the same program id so there isnt a way to discriminate between them otherwise

buffalojoec: Ahh I see, so the main purpose being to overwrite the stake program in the second one. Yeah, I think you've got it right then.

2501babe:
the branch rename seems to have dropped this prs connection to ci so im gonna merge it as-is once i can switch from local mollusk and then see what needs fixed in-flight. i think ci is still just fmt/clippy and doesnt run tests so ill have to add some workflow files anyway

program/tests/stake_instruction.rs
```diff
@@ -0,0 +1,7054 @@
...
+        &instruction::move_lamports(
+            &Pubkey::new_unique(),
+            &Pubkey::new_unique(),
+            &Pubkey::new_unique(),
+            100,
+        ),
+        Err(ProgramError::InvalidAccountData),
+    );
+}
+
```
joncinque, line 451: The test `test_spoofed_stake_accounts` doesn't seem to be here -- is that intentional?

joncinque (commented):
Sorry for the super lateness on this, but it looks good! I mainly looked at the setup, test names, and usage of `is_bpf()` as you suggested, without reading the logic deeply.

There's just the one missing test that I noticed.

Otherwise, will you re-add the tests from https://github.com/anza-xyz/agave/blob/master/programs/stake/src/stake_state.rs in future work? Thankfully, I think it'll be much simpler than this one

program/tests/stake_instruction.rs
2501babe: yes, that test tries to trigger an invalid owner error for every instruction if the first account isnt owned by the program. we dont get the same errors most of the time because this is just an artifact of irregular handling: the first line in every native stake program instruction handler is `let me = get_stake_account()?;` which overrides all other errors like signer checks, checks on other accounts, etc

joncinque: Gotcha, makes sense, thanks!

PR #26, opened 2024-12-14
Interface tests

2501babe:
this adds tests that attempt to exhaustively check the interfaces of every stake instruction, to make up for gaps in coverage from the existing tests, particularly with lockups

### DIFF
--- a/program/tests/interface.rs
+++ b/program/tests/interface.rs
@@ -3,7 +3,7 @@
 use {
     arbitrary::{Arbitrary, Unstructured},
     mollusk_svm::{result::Check, Mollusk},
-    solana_account::{Account, AccountSharedData, ReadableAccount, WritableAccount},
+    solana_account::{Account, ReadableAccount, WritableAccount},
     solana_sdk::{
         instruction::{AccountMeta, Instruction},
         native_token::LAMPORTS_PER_SOL,
@@ -206,7 +206,7 @@ impl Env {
 
     // get the accounts from our account store that this transaction expects to see
     // we dont need implicit sysvars, mollusk resolves them internally via syscall stub
-    fn resolve_accounts(&self, account_metas: &[AccountMeta]) -> Vec<(Pubkey, AccountSharedData)> {
+    fn resolve_accounts(&self, account_metas: &[AccountMeta]) -> Vec<(Pubkey, Account)> {
         let mut accounts = vec![];
         for account_meta in account_metas {
             let key = account_meta.pubkey;
@@ -234,7 +234,7 @@ impl Env {
             } else if let Some(account) = self.base_accounts.get(&key).cloned() {
                 account.into()
             } else {
-                AccountSharedData::default()
+                Account::default()
             };
 
             accounts.push((key, account_shared_data));
--- a/program/tests/interface.rs
+++ b/program/tests/interface.rs
@@ -230,11 +230,9 @@ impl Env {
                     .keyed_account_for_stake_history_sysvar()
                     .1
             } else if let Some(account) = self.override_accounts.get(&key).cloned() {
-                account.into()
-            } else if let Some(account) = self.base_accounts.get(&key).cloned() {
-                account.into()
+                account
             } else {
-                Account::default()
+                self.base_accounts.get(&key).cloned().unwrap_or_default()
             };
 
             accounts.push((key, account_shared_data));
@@ -313,7 +311,7 @@ impl StakeInterface {
             Self::Authorize(checked, authority_type, lockup_state) => {
                 env.update_stake(
                     &STAKE_ACCOUNT_BLACK,
-                    &not_just_stake(
+                    &initialized_stake(
                         STAKE_ACCOUNT_BLACK,
                         minimum_delegation,
                         false,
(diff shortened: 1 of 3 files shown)


program/tests/interface.rs
```diff
@@ -0,0 +1,832 @@
...
+) -> StakeStateV2 {
+    i_cant_believe_its_not_stake(
+        Pubkey::default(),
+        stake_pubkey,
+        stake,
+        StakeStatus::Initialized,
+        common_authority,
+        lockup,
+    )
+}
```
2501babe, line 724: im so sad professionalism demands i rename these

2501babe:
@joncinque this is low priority so feel free to delay it if youre busy, i just wanted to get it in review now that it points at tip of master. im probably going to have a couple more important prs related to simd170 next week

program/tests/interface.rs
```diff
@@ -0,0 +1,830 @@
...
+    std::{
+        collections::{HashMap, HashSet},
+        sync::LazyLock,
+    },
+};
+
+// NOTE ideas for future tests:
+// * fail with different vote accounts on operations that require them to match
+// * fail with different authorities on operations that require them to match
+// * adding/changing lockups to ensure we always fail when violating lockup
```
joncinque, line 40: I really appreciate the full coverage of these tests, but can you add some information about how this works? I'm pretty sure I'll get confused every single time I go back to read this.

Feel free to rewrite it to your liking, but I think something like this would help, but maybe I'm just a bit slow sometimes:

> `StakeInterface` encapsulates every possible variation on instructions that we want to test, so that we can use random data to generate valid instructions.

> All of the constants and addresses in this file are used with mollusk to setup base usable accounts, which `StakeInterface` then puts into a suitable state for particular instructions. For example, `merge` sets up two stake accounts with appropriate lockups and authorities.

> The `Env` also contains a set of `override_accounts`, which are used by `StakeInterface` to update stake accounts for a particular instruction.

program/tests/interface.rs
```diff
@@ -0,0 +1,830 @@
...
+    Authorize(bool, AuthorityType, LockupState),
+    AuthorizeWithSeed(bool, AuthorityType, LockupState),
+    SetLockup(bool, LockupState, LockupState),
+    DelegateStake(LockupState),
+    Split(LockupState, bool),
+    Merge(LockupState),
+    MoveStake(LockupState, bool, bool),
+    MoveLamports(LockupState, bool, Option<bool>),
+    Withdraw(Option<bool>, LockupState, bool),
+    Deactivate(LockupState),
```
joncinque, line 281: nit: For readability, can you change the data from tuples to structs? ie

```
Authorize {
    checked: bool,
    authority_type: AuthorityType,
    lockup_state: LockupState
}
```

program/tests/interface.rs
```diff
@@ -0,0 +1,830 @@
...
+    InitializeChecked,
+    Authorize(bool, AuthorityType, LockupState),
+    AuthorizeWithSeed(bool, AuthorityType, LockupState),
+    SetLockup(bool, LockupState, LockupState),
+    DelegateStake(LockupState),
+    Split(LockupState, bool),
+    Merge(LockupState),
+    MoveStake(LockupState, bool, bool),
+    MoveLamports(LockupState, bool, Option<bool>),
+    Withdraw(Option<bool>, LockupState, bool),
```
joncinque, line 280: nit: The `Option<bool>`s here made me particularly confused -- can this just be an enum like `StakeStatus`?

program/tests/interface.rs
```diff
@@ -0,0 +1,830 @@
...
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Arbitrary)]
+enum StakeStatus {
+    Uninitialized,
+    Initialized,
+    Activating,
+    Active,
+    Deactivating,
+    Deactive,
+}
```
joncinque, line 647: Doesn't need to be done in this PR, but we could consider adding the `Arbitrary` derivation on the types within the interface crate with a `dev-context-only-utils` feature

joncinque (commented):
Looks great overall! Most of my comments are minor to help us in the future whenever we have to revisit this

program/tests/interface.rs
2501babe: will do! but to explain my thinking, `to_instruction()` is the only place where i felt these terms "really" needed to be defined, the `StakeInterface` struct itself is just the types needed for `Arbitrary` to generate all valid values (also the reason for `Option<bool>` as just "an arbitrary ternary value")

program/tests/interface.rs
2501babe: im not sure what i would use them for, there are so many external data dependencies (authorities on concrete keypairs, lockups on keypairs and clock, activation/deactivation on clock, delegation and rent exempt reserve on account lamports), invalid states, and valid states that are astronomically unlikely for a randomly generated struct to hit. i wrote my own so that `Arbitrary` would cover all important valid, and no invalid, states

joncinque: Definitely, I don't mean to do it for all of them, for the exact reason you outlined. But wherever possible, it's nice to avoid type duplication

joncinque (approved):
Looks great, thanks!

PR #27, opened 2024-12-19
use bpf-entrypoint feature

buffalojoec:
The rest of the programs we've written in BPF have swapped `no-entrypoint` for `bpf-entrypoint`, where the feature is now _additive_ to enable the program ELF's entrypoint binding.

This change brings the Stake program into that same pattern.

### DIFF
--- a/program/Cargo.toml
+++ b/program/Cargo.toml
@@ -8,7 +8,7 @@ license = "Apache-2.0"
 edition = "2021"
 
 [features]
-no-entrypoint = []
+bpf-entrypoint = []
 test-sbf = []
 
 [dependencies]
--- a/program/src/entrypoint.rs
+++ b/program/src/entrypoint.rs
@@ -1,7 +1,5 @@
 //! Program entrypoint
 
-#![cfg(not(feature = "no-entrypoint"))]
-
 use {
     crate::processor::Processor,
     solana_program::{account_info::AccountInfo, entrypoint::ProgramResult, msg, pubkey::Pubkey},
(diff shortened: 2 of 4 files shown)


febo (approved):
Looks good. One little comment: we will need to build the program with the `bpf-entrypoint` feature enabled so the clients tests work:

* https://github.com/solana-program/stake/blob/main/.github/workflows/main.yml#L79

 At the moment it is not an issue since we don't have clients tests, so we could do it either in this PR or in a future one.

2501babe:
ty for doing this, i was aware of the feature switch but it slipped my mind

maybe unrelated to this pr but is there a better ux for the new paradigm than always supplying `--features bpf-entrypoint` manually? i cloned this to make sure tests run, and i have to provide the flag even to `cargo test-sbf`. seems a bit annoying

2501babe (approved):

buffalojoec:
> maybe unrelated to this pr but is there a better ux for the new paradigm than always supplying `--features bpf-entrypoint` manually? i cloned this to make sure tests run, and i have to provide the flag even to `cargo test-sbf`. seems a bit annoying that it doesnt do the right thing automatically

I was considering adding it to the `build-sbf` and `test-sbf` commands, but I wasn't totally sure if we wanted to make it a global standard for the platform tools. It could just be optionally enabled for crates that specify the feature. Wdyt?

2501babe:
yea if added it would need to check if the feature exists in Cargo.tom because if you specify the flag and the feature doesnt exist then cargo will halt with an error

joncinque (commented):
This looks good to me! Can you also add the feature to `scripts/program/build.mjs`?

PR #33, opened 2025-01-17
interface: impl SysvarId for StakeHistory

kevinheavey:
This trait impl needs to be in either solana-stake-interface or solana-sysvar-id because of the orphan rule. The interface crate is the obvious choice because it would add a bunch of deps to solana-sysvar-id

### DIFF
--- a/interface/src/stake_history.rs
+++ b/interface/src/stake_history.rs
@@ -3,7 +3,7 @@
 //! [sv]: https://docs.solanalabs.com/runtime/sysvars#stakehistory
 
 pub use solana_clock::Epoch;
-use std::ops::Deref;
+use {solana_sysvar_id::declare_sysvar_id, std::ops::Deref};
 
 pub const MAX_ENTRIES: usize = 512; // it should never take as many as 512 epochs to warm up or cool down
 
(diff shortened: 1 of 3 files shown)


interface/src/stake_history.rs
```diff
@@ -65,6 +65,8 @@ impl std::ops::Add for StakeHistoryEntry {
 #[derive(Debug, PartialEq, Eq, Default, Clone)]
 pub struct StakeHistory(Vec<(Epoch, StakeHistoryEntry)>);
 
+declare_sysvar_id!("SysvarStakeHistory1111111111111111111111111", StakeHistory);
```
joncinque, line 68: Elsewhere we're re-exporting whatever's declared in solana-sdk-ids, but since this is outside the SDK, I think it makes more sense to redefine it.

joncinque (approved):
Looks great, thanks!

interface/src/stake_history.rs
kevinheavey: Oh I would have preferred to just re-export from solana-sdk-ids but it's not published yet

joncinque: I've been doing some prereleases of all the new crates, so https://crates.io/crates/solana-sdk-ids/0.0.3 probably has what you want :eyes:

PR #34, opened 2025-01-17
fix bincode feature

kevinheavey:
Currently the interface crate doesn't compile with `cargo build -F bincode` because it's missing `solana-system-interface/bincode`


febo (approved):
Thanks! I will get CI updated on a separate PR so we can avoid missing features.

PR #38, opened 2025-01-23
scripts: Update repository scripts

febo:
This PR updates the scripts of the repository. It uses the "refactored" ones similar to the [System](https://github.com/solana-program/system) repository.


scripts/rust.mts
```diff
@@ -0,0 +1,162 @@
...
+  );
+}
+
+async function lintClippy() {
+  return cargo(
+    getToolchainArgument('lint'),
+    'clippy',
+    popArgument(args, '--fix') ? ['--fix'] : []
+  );
+}
```
joncinque, line 59: Doesn't need to be done in this PR, but we typically provide some more arguments to clippy, ie. `cargo +nightly-2024-11-22 clippy --all-targets --all-features -- --deny=warnings --deny=clippy::default_trait_access --deny=clippy::arithmetic_side_effects --deny=clippy::manual_let_else --deny=clippy::used_underscore_binding`

package.json
```diff
@@ -1,34 +1,48 @@
...
+    "js:lint": "tsx ./scripts/js.mts lint clients/js",
+    "js:publish": "tsx ./scripts/js.mts publish clients/js",
+    "js:test": "tsx ./scripts/js.mts test clients/js",
+    "rust:format": "tsx ./scripts/rust.mts format clients/rust",
+    "rust:lint": "tsx ./scripts/rust.mts lint clients/rust",
+    "rust:lint:clippy": "tsx ./scripts/rust.mts lint-clippy clients/rust",
+    "rust:lint:docs": "tsx ./scripts/rust.mts lint-docs clients/rust",
+    "rust:lint:features": "tsx ./scripts/rust.mts lint-features clients/rust",
+    "rust:publish": "tsx ./scripts/rust.mts publish clients/rust",
+    "rust:test": "tsx ./scripts/rust.mts test clients/rust",
```
joncinque, line 28: Is there a reason that these aren't prefixed with `clients`? Ie. `clients:js:format`? The reason I ask is that it'll be inconsistent with the other repos

package.json
```diff
@@ -1,34 +1,48 @@
...
-    "clients:rust:publish": "zx ./scripts/client/publish-rust.mjs",
-    "clients:rust:test": "zx ./scripts/client/test-rust.mjs"
+    "validator:stop": "tsx ./scripts/helpers/stop-validator.mts",
+    "program:build": "tsx ./scripts/rust.mts build-sbf program --features bpf-entrypoint",
+    "program:format": "tsx ./scripts/rust.mts format program",
+    "program:lint": "tsx ./scripts/rust.mts lint program",
+    "program:lint:clippy": "tsx ./scripts/rust.mts lint-clippy program",
+    "program:lint:docs": "tsx ./scripts/rust.mts lint-docs program",
+    "program:lint:features": "tsx ./scripts/rust.mts lint-features program",
+    "program:test": "tsx ./scripts/rust.mts test program",
```
joncinque, line 17: Is there a reason this is inconsistent with the other repos which prefix with `programs` instead of `program`? I'm asking because there are monorepo jobs which test downstream repos, and it's expecting to be able to do `programs:build` in all them, for example

joncinque (approved):
Just some small things, but those can get fixed later. Looks good!

scripts/rust.mts
febo: I try switching from `nightly-2024-08-08` to `nightly-2024-11-22` and we already get warnings on both program and client crates – I think it is better to fix those in a different PR.

package.json
febo: I think this was the new "template" that we agreed, dropping the "client". Only the system repository is following this at the moment, so we could revert this pattern.

package.json
febo: Same here, I think we renamed to `program` in the last refactoring since we usually have a single program per repository.

package.json
lorisleiva: Hmm yeah I think this is something that wasn't obvious with the `system` repo because the program doesn't live in the repo but I agree with Jon, this lack consistency and isn't predictable.

I really like our latest refactoring where, for instance, everything Rust-related is in its own script and then you have sub commands within that script.

Maybe we should lean into that more in the pnpm scripts as well. We could have a simple `rust` script and let the user of that script to provide the arguments.

So instead of something like `pnpm program:format` we would do `pnpm rust format program`. That way, there are less "indirections" between the scripts and how they are invoked. What do you think?

joncinque: The simplicity of running the commands is definitely nice, but on the flip-side you lose the "self-documenting" nature of package.json, which currently tells you everything that you should be able to run successfully in a repo.

I don't have a very strong preference, but I would stick with the `pnpm program:format` solely for self-documentation.

joncinque (approved):

PR #39, opened 2025-01-23
ci: Refactor publish rust crate

febo:
This PR refactors the `publish-rust-client` to allow publishing rust crates for the repository (e.g., `interface`, `clients/rust`).

### DIFF
--- a/.github/workflows/publish-rust-crate.yml
+++ b/.github/workflows/publish-rust-crate.yml
@@ -37,7 +37,7 @@ on:
         default: true
 
 jobs:
-  test_interface:
+  test:
     name: Test Rust Crate
     runs-on: ubuntu-latest
     steps:
@@ -64,10 +64,10 @@ jobs:
       - name: Test
         run: tsx ./scripts/rust.mts test "${{ inputs.crate_path }}"
 
-  publish_rust:
+  publish:
     name: Publish Rust Crate
     runs-on: ubuntu-latest
-    needs: test_interface
+    needs: test
     permissions:
       contents: write
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-rust-crate.yml
```diff
@@ -33,8 +37,8 @@ on:
         default: true
 
 jobs:
-  test_rust:
-    name: Test Rust client
+  test_interface:
```
joncinque, line 40: nit: maybe just call this `test`?

.github/workflows/publish-rust-crate.yml
```diff
@@ -92,8 +95,8 @@ jobs:
 
       - name: Set Git Author
         run: |
-          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
-          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions@github.com"
+          git config --global user.name "github-actions"
```
joncinque, line 99: Is this change needed?

.github/workflows/publish-rust-crate.yml
```diff
@@ -122,4 +125,4 @@ jobs:
         if: github.event.inputs.create_release == 'true' && github.event.inputs.dry_run != 'true'
         uses: ncipollo/release-action@v1
         with:
-          tag: rust@v${{ steps.publish.outputs.new_version }}
+          tag: ${{ steps.publish.outputs.crate_name }}@v${{ steps.publish.outputs.new_version }}
```
joncinque, line 128: nit: The crate name is a bit ugly for the tag... can we use the end of the path? Most of the other tags are using that, ie. `js` / `interface` / `program`

joncinque (commented):
A couple of things on my side, the rest looks good!

.github/workflows/publish-rust-crate.yml
febo: It should avoid seeing the `github-actions[bot]` as a contributor.

joncinque: Why do we want to avoid seeing `github-actions[bot]` as a contributor? I thought that was a good thing so that commits are associated with real github users :sweat_smile:

joncinque (approved):
Looks good! We can revisit the github actions user bit later

PR #42, opened 2025-01-28
ci: Fix publish workflow

febo:
### Problem

The "publish" workflow is missing a `npx` command in order to run the `tsx` scripts; additionally, the `build` toolchain is not being set.

### Solution

Add the missing command and toolchain definition.

A "dry-run" of the workflow now works: https://github.com/solana-program/stake/actions/runs/13001907231

### DIFF
--- a/.github/workflows/publish-rust-crate.yml
+++ b/.github/workflows/publish-rust-crate.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       crate_path:
-        description: Path to directory with crate to release
+        description: Path to crate
         required: true
         type: string
       level:
--- a/.github/workflows/publish-rust-crate.yml
+++ b/.github/workflows/publish-rust-crate.yml
@@ -53,16 +53,16 @@ jobs:
           solana: true
 
       - name: Format
-        run: npx tsx ./scripts/rust.mts format "${{ inputs.crate_path }}"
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-rust-crate.yml
```diff
@@ -49,20 +49,20 @@ jobs:
         with:
           cargo-cache-key: cargo-test-publish-${{ inputs.crate_path }}
           cargo-cache-fallback-key: cargo-test-publish
-          toolchain: format, lint, test
+          toolchain: build, format, lint, test
           solana: true
 
       - name: Format
-        run: tsx ./scripts/rust.mts format "${{ inputs.crate_path }}"
+        run: npx tsx ./scripts/rust.mts format "${{ inputs.crate_path }}"
```
lorisleiva, line 56: `npx` isn't ideal because IIRC it will download the package on-demand even though it is actually part of the dependencies.

How would you feel about having a `pnpm rust` script instead? So here we'd be able to do:

```
run: pnpm rust format "${{ inputs.crate_path }}"
```

febo: I replaced `npx` by `pnpm` – would that work better? Perhaps it would be nicer not to add another script on `package.json` that only CI would use.

lorisleiva (approved):

PR #45, opened 2025-01-30
publish: Fix workflow

febo:
#### Problem

While the `publish-rust-crate` can successfully publish a crate, it fails to create a git tag and GitHub release due to using the wrong path to look up the `Cargo.toml` file.

#### Solution

Fix the `Cargo.toml` look up by using the correct value. It also fixes the docs link on the `README.md` of the interface crate.

### DIFF
--- a/interface/README.md
+++ b/interface/README.md
@@ -22,4 +22,4 @@ This will add the `solana-stake-interface` dependency with the `bincode` feature
 
 ## Documentation
 
-Read more about the Stake program on the crate [documentation](https://docs.rs/solana-stake-interface/latest/solana-stake-interface/).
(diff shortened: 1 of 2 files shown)


febo:
> Sorry, but how does this fix it? `manifestPath` already contains `Cargo.toml` and then `getCargo` again appends `Cargo.toml` at
> 
> https://github.com/solana-program/stake/blob/bb6932ed816eb39205102eee2e0cbc0cd511dcaa/scripts/helpers/utils.mts#L80

You are right! The correct value to use here is `libraryPath`.

joncinque (approved):
Makes much more sense, thanks!

PR #48, opened 2025-02-05
publish: Fix crate name parsing

febo:
#### Problem

Currently there is a problem in how the crate name parsing is done on the publish crate workflow – it fails to determine the correct path for the crate, which prevents the script to parse the name and version of the crate.

#### Solution

Modify the `parseCliArguments` to return the original `relativePath` value so that the script can parse the name and version of the crate.


lorisleiva (approved):
Perfect! 🍺

PR #49, opened 2025-02-07
update versions, fix mollusk tests

2501babe:
bpf stake program presently does not build due to versioning issues. this relaxes the solana version locks in `dev-dependencies` (necessary because of matching version locks in mollusk 0.0.13 which have themselves been relaxed in 0.0.15) and fixes some things to update mollusk versions

mollusk wants `Account` rather than `AccountSharedData` now so rather than a gigantic changelog reworking all the tests i just covert on ingress and egress

unrelated note but the `ProgramTest` tests are all broken due to [simd170](https://github.com/solana-foundation/solana-improvement-documents/pull/170) restricting native programs to 3k cus. there is some machinery i will have to look at to see how to make `ProgramTest` tell the cost model our bpf program is no longer "native"

once we fix the tests we need to set up ci properly and hopefully get looped into monorepo ci to prevent breakage in the future

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2363,28 +2363,29 @@ dependencies = [
 
 [[package]]
 name = "mollusk-svm"
-version = "0.0.13"
+version = "0.0.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1fceaf67fe3f95a9478f4f5b0d71e77c073eee7a795a74d6143317a22454c289"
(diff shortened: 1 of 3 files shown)


joncinque (approved):
Beautiful, thanks!

And yikes, the reserved CUs work is a bit scary. Do we need to get a fix in for that asap or backport something to 2.1 to accommodate the native programs in BPF?

PR #51, opened 2025-02-10
program: add sbf tests to ci

2501babe:


febo (approved):
Looks good! We can check if there is a way to reuse the `pnpm` script in a separate PR.

2501babe:
yea to be clear i tested the pnpm command and it does work. the problem is if you have a test failure, it prints a generic node error and a json struct with a truncated version of only the first dozen or so lines of the rust execution logs. i think for rust tests we should just call cargo directly

febo:
> yea to be clear i tested the pnpm command and it does work. the problem is if you have a test failure, it prints a generic node error and a json struct with a truncated version of only the first dozen or so lines of the rust execution logs. i think for rust tests we should just call cargo directly

Ah, yeah, the output is a bit messy. I think the cargo error gets printed before the node error, so the error will appear twice. I tried to stop `pnpm` to showing an error, but then it does not exit with an error code. I will see if there are any other tweaks we can do.

PR #50, opened 2025-02-10
program: disable minimum cus for program tests

2501babe:

PR #52, opened 2025-02-11
program: update to 2.1.10

2501babe:
solana_program 2.1.10 is a hard minimum requirement due to work done on builtin compute usage

this completes the work started by #49 and #50 to fix breakage from upstream that went unnoticed. #51 (yet to land) will start running program tests in ci

when the Great Monorepo Cleftishment is complete we should start thinking about adding the bpf programs as downstream repos, because if this happens again it would be because we went from "ci doesnt run tests" to "ci doesnt run often enough because we dont need to make changes to the code"

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -65,9 +65,9 @@ dependencies = [
 
 [[package]]
 name = "agave-transaction-view"
-version = "2.1.9"
+version = "2.1.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


joncinque (approved):
Looks great! That's also a good point about adding downstream testing for the BPF programs, it should be pretty easy to add them at https://github.com/anza-xyz/agave/blob/bbb7e5fd153044045f611e70ddb1b65dc068ab6b/.github/workflows/downstream-project-spl.yml#L46 as long as they follow the same general commands

PR #54, opened 2025-02-27
IDL: Update discriminants to u32 instead of u8

joncinque:
#### Problem

The IDL used to generate the stake program client declares the instruction discriminant as a `u8`, when the program actually uses `u32`s for the discriminant.

#### Summary of changes

Update the IDL, update the generate script, then regenerate all of the clients.

NOTE: This doesn't cover the `StakeState` and `StakeStateV2` enums, which both use a `u32` discriminant too. I wasn't sure how to do that in the IDL, so I omitted it.


joncinque:
Looks like there some codegen messing up for Rust -- do I need to update something? I tried updating all of the codama deps, but it didn't resolve the issue.

I then removed the trait overrides from the generate script, but that caused other issues because of `f64`s in the structs, but I'm not sure how to get around that. After that, I just need to update the hooked file. So once the f64 thing is resolved, this should be ok

lorisleiva:
It looks like the new default traits for Rust items are causing an issue in this upgrade. You could use the following base traits to avoid using `Eq` on all items.

```ts
const traitOptions = {
    baseDefaults: [
        'borsh::BorshSerialize',
        'borsh::BorshDeserialize',
        'serde::Serialize',
        'serde::Deserialize',
        'Clone',
        'Debug',
        // 'Eq', <- Remove 'Eq' from the default traits.
        'PartialEq',
    ],
};
```

You can also be more granular with this if you wanted to ([see option docs](https://github.com/codama-idl/codama/tree/main/packages/renderers-rust#trait-options)).

Regarding updating the `StakeState` and `StakeStateV2` enums, I don't believe you can do that on the Anchor IDL (which is what we currently have for that program) but you can use a transformer visitor on the `generate-clients` script to adjust the tree before rendering.

[2min later]

Looking at the current script though it looks like something like this already exists. Is is not materialising as a u32 in the generated code?

```ts
codama.update(
  c.bottomUpTransformerVisitor([
    {
      // enum discriminator -> u32
      select: '[definedTypeNode]stakeStateV2.[enumTypeNode]',
      transform: (node) => {
        c.assertIsNode(node, 'enumTypeNode');
        return {
          ...node,
          size: c.numberTypeNode('u32'),
        };
      },
    },
  ])
);
```

joncinque:
> It looks like the new default traits for Rust items are causing an issue in this upgrade. You could use the following base traits to avoid using Eq on all items.

Ah ok great, done!

> Looking at the current script though it looks like something like this already exists. Is is not materialising as a u32 in the generated code?

You're right, it's correct for `StakeStateV2`, but not `StakeState`, so I added it for `StakeState` as well

joncinque:
Oh, but note that the u32 enum discriminant on structs isn't possible with Rust, since it's just using the normal derive macro for borsh

lorisleiva:
Oh yeah. We really need Rust codecs so anything doable in JS can be done in Rust. Aside from that is this okay to merge? I'm gonna rebase my `@solana/kit` upgrade on top of that afterwards.

joncinque:
Yeah it's ready from my side! So if you approve, I'll merge

lorisleiva (approved):
Thanks!

Issue #53, opened 2025-02-11
remove all sysvar pubkeys from instructions interface


2501babe:
because of syscalls, we do not need to pass in _any_ sysvar accounts. we can also get rid of stake config entirely. to preserve the interface for past transactions, we can skip the relevant accounts in a wrapper on `next_account_info()`

this should (in the case of clock and rent) and absolutely will greatly (in the case of stake history) reduce CU usage because these accounts do not need to be parsed into `AccountInfo`s. it will also make transactions that use stake account instructions typically 32-128 bytes smaller

this was not done for the initial version on purpose because above all else except correctness, we prioritized following the native stake program as closely as possible, to de-risk the native/bpf switchover. but this is probably the first pr i will land after launch

PR #55, opened 2025-03-03
interface: remove stake config

2501babe:
now that mainnet-beta is on 2.1, this can be removed. this is a nonbreaking change: the native stake and bpf stake versions of `DelegateStake` both gracefully handle the presence, or absence, of stake config

### DIFF
--- a/interface/src/instruction.rs
+++ b/interface/src/instruction.rs
@@ -10,7 +10,7 @@ use {
 };
 #[cfg(feature = "bincode")]
 use {
-    crate::{config, program::ID, state::StakeStateV2},
+    crate::{program::ID, state::StakeStateV2},
     solana_instruction::{AccountMeta, Instruction},
 };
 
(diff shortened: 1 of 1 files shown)


joncinque (commented):
If we remove the account in the instruction creator, we're relying on extracting the stake authority as the stake config at https://github.com/solana-program/stake/blob/74fa890dabce6c585c1038767fbf8802816d4dbe/program/src/processor.rs#L367-L372

which we can likely break unwittingly in the future.

Can you also remove the extracting the account in the program with a big huge comment saying that old instruction creators may be providing the stake config and to never change the logic to add new accounts?

2501babe:
if you think this change is nontrivial id rather close this pr and reopen when we do #53 than edit the program, since its in audit now. my thinking was that since we are only editing the instruction builder, we know that we provide the authority, because its part of the instruction builder

joncinque:
I'm thinking it might be better to change all the sysvars at once, that way we're sure to not mess anything up.

PR #62, opened 2025-03-07
deps: Bump Solana deps to v2.2

joncinque:
#### Problem

The Solana 2.2 crates and tools are available, but the repo is still using 2.1.

#### Summary of changes

Pretty simple, but bump up the sdk crates to v2.2.1, non-sdk crates to v2.2.0, toolchain to 1.84.1, and nightly to 2024-11-22. Also add the workspace lints.


2501babe:
oh wait are we doing this now? i just mentioned offhand in another issue that i assumed we were holding at 2.1 until launch

joncinque:
Ah ok, I can hold off if you prefer. I mainly want the stake interface bumped because it's still on older versions, which is causing things like semver checks to fail

2501babe:
im not strongly opposed, just making sure we are on the same page! its probably fine since it requires no code changes

febo:
In practice we don't need to bump the interface and client dependencies, since `^2.1` would be compatible with `2.2`. Is the idea to always move the dependencies to the latest? The drawback is that when we introduce a fix to one of these crates, then users will also need to bump their dependencies to the latest to get the fix.

joncinque:
I mainly want to get everything on the v2.2.1 sdk crates so that everything can benefit from the versions with relaxed dependencies. I'm hoping that the change can even help us remove the patches from the sdk repo.

febo (dismissed):

febo:
I think it only needs a rebase.

PR #68, opened 2025-05-07
fix: parseWithdrawInstruction if statement

emi-tangany:
Seems like the `lockupAuthority` is optional, which means that the number of accounts available must be strictly 5 or 6 otherwise it should throw an error. Please let me know if I may be missing something because the previous version was throwing an error when 5 accounts were available and that should have been a valid use case.

If the change is accepted, when will the next version of the library be released?

### DIFF
--- a/clients/js/src/generated/instructions/withdraw.ts
+++ b/clients/js/src/generated/instructions/withdraw.ts
@@ -247,9 +247,9 @@ export function parseWithdrawInstruction<
     IInstructionWithAccounts<TAccountMetas> &
     IInstructionWithData<Uint8Array>
 ): ParsedWithdrawInstruction<TProgram, TAccountMetas> {
-  if (instruction.accounts.length < 6) {
+  if (instruction.accounts.length < 5 || instruction.accounts.length > 6) {
     // TODO: Coded error.
-    throw new Error('Not enough accounts');
+    throw new Error('Withdraw Instruction expects 5 or 6 accounts');
   }
   let accountIndex = 0;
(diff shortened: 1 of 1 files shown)


lorisleiva (changes requested):
Unfortunately, if you look at the top of this file, you will see a comment saying it is generated and cannot be updated directly. If the 7th account is optional, it should be marked as such on the Codama IDL so this helper function can be generated properly.

emi-tangany:
Thanks for the explanation @lorisleiva, I have not seen the top of the file before 👍 
Based on the following [line](https://github.com/solana-program/stake/pull/68/files#diff-c77057c9a7b09dfb8a23db0f00c160c57cf22e9748b25cededcdebcdd790cdc2R260-R265) that states that the property is optional I would assume that this auto-generated code is wrongly generated, right?

lorisleiva:
Oh I see! It seems it is actually an issue with the optional account strategy used by the instructions of this program. I've proposed a fix for this in https://github.com/solana-program/stake/pull/69.

emi-tangany:
Thanks for that fix; it is very much appreciated! Would you mind posting the link to the new version when published in NPM ? then I guess we could close this PR 🙏

lorisleiva:
[Version `0.2.1` published](https://www.npmjs.com/package/@solana-program/stake/v/0.2.1)! Sorry for the delay.

emi-tangany:
Thanks so much @lorisleiva, very much appreciated!

PR #69, opened 2025-05-07
Use omitted optional account strategy on IDL

lorisleiva:
When converting an Anchor IDL to a Codama IDL, the adapter defaults to using the `"programId"` optional account strategy for all instructions. This allows positional optional accounts as introduced in Anchor which means when an account is not provided, it is replaced by the program ID to only used a single byte in the transaction.

However, this is not the appropriate optional account strategy for the stake account as it simply omits optional accounts. Therefore, this PR replaces the `optionalAccountStrategy` of all instructions to the `"omitted"` strategy.

This in turns updates the generated code to be more accurate when creating and parsing instructions.

Note that, I've had to adjust the `generate:clients` script slightly so it calls `pnpm format:fix` because the Prettier used post-generation is slightly different now then the one we use in the `pnpm format` script. I'm not too sure why though, they both should use the same prettier configs but at least this brings us back to green on CI.

### DIFF
--- a/clients/js/src/generated/instructions/authorize.ts
+++ b/clients/js/src/generated/instructions/authorize.ts
@@ -51,7 +51,10 @@ export type AuthorizeInstruction<
     | string
     | IAccountMeta<string> = 'SysvarC1ock11111111111111111111111111111111',
   TAccountAuthority extends string | IAccountMeta<string> = string,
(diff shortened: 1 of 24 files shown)


joncinque (approved):
Looks good to me!

lorisleiva:
Sorry I hadn't realised this wasn't merged.

PR #70, opened 2025-06-18
`StakeInstructionBuilder` as replacement for individual fns

grod220:
Proposal to move the stake program's instruction building to a builder pattern allowing for more flexibility w/ variants.

```rust
StakeInstructionBuilder::initialize(&stake_pubkey, &authorized, &lockup)
  .create_account(&from_pubkey, lamports)
  .build();

StakeInstructionBuilder::split(&stake_pubkey, &stake_authority_pubkey, &split_stake_pubkey, split_lamports)
  .with_allocate_with_seed(&base, seed)
  .build()
```

== In progress ==

### DIFF
--- a/interface/src/instruction.rs
+++ b/interface/src/instruction.rs
@@ -329,6 +329,7 @@ pub struct AuthorizeCheckedWithSeedArgs {
     pub authority_owner: Pubkey,
 }
 
+#[deprecated(since = "1.2.0", note = "Please use `StakeInstructionBuilder` instead")]
 #[cfg(feature = "bincode")]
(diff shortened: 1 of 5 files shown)


joncinque (commented):
The auto-generated rust clients with Codama include builders for instruction creators, so we'd probably want to have these auto-generated instead of including them in the interface crate, which is meant to be just the most low-level stuff.

grod220:
Helpful! Looks like the route to improving the helpers is then: codama-rs finalization -> integrate macros into program -> generate IDL -> generate clients from IDL -> publish crate with updated builders.

PR #78, opened 2025-08-06
test epoch rewards blocks program usage

2501babe:
this is part of removing the stake program from agave (https://github.com/anza-xyz/agave/pull/7203). right now, the agave repo still gets its version of the stake program from `programs/stake/` for all tests, which is not just useless but actually dangerous. adding the stake program to `TestValidator` is easy, but some tests that use a raw `Bank` are more tricky to deal with

we have to decide which things should:
1. stub out the stake program (because their usage of stake is incidental)
2. move to this repo (because they essentially just test the stake program)
3. need to use the binary blob stake program being added to agave (because they test a fundamental stake program and validator interaction)

this test is intended to replace `test_program_execution_restricted_for_stake_account_in_reward_period()` in `runtime/src/bank/partitioned_epoch_rewards/mod.rs`, which you should look at while reviewing this pr to see if you agree. it does a lot of bank/tower/vote things to naturally advance the bank into and out of the rewards period. in my view this is unecessary and all the test really cares about is that the stake program aborts, and other tests cover the rewards period itself. if youre of the same view ill delete that test in its entirety

### DIFF
--- a/program/tests/interface.rs
+++ b/program/tests/interface.rs
@@ -901,6 +901,7 @@ fn fully_configurable_stake(
     }
 }
 
+// test all unmodified transactions succeed, to ensure other tests test what they purport to test
 #[test]
 fn test_all_success() {
     let mut env = Env::init();
@@ -912,6 +913,7 @@ fn test_all_success() {
     }
 }
 
+// all signers are essential; missing any one signer is a fail
 #[test]
 fn test_no_signer_bypass() {
     let mut env = Env::init();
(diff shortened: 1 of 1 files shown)


joncinque (approved):
This test looks great!

I'm almost totally onboard with removing `test_program_execution_restricted_for_stake_account_in_reward_period`.

My only concern is that the test *also* checks that it's possible to credit a stake account during reward payout, which could be useful in case we mistakenly decide to restrict that somehow in the future.

So I would remove the stake interactions in that test, but keep the crediting. What do you think?

PR #82, opened 2025-08-13
program: reallocate stake account on zero lamports

### LINKED ISSUE #71: program: Properly delete accounts
#### Problem

When a stake account is "deleted", through being merged or withdrawn from, we just set the state back to `Uninitialized`, which leaves the account not completely deleted. Most protocols realloc the account to a size of 0 and assign its owner back to the system program.

#### Proposed Solution

Get the stake program in line with other programs, and completely delete accounts.

2501babe:
instead of zeroing account data, we should realloc accounts to 0 size, to ease headaches of downstream programs dealing with stake accounts across multiple instructions in the same transaction

closes #71

### DIFF
--- a/program/tests/interface.rs
+++ b/program/tests/interface.rs
@@ -952,3 +952,45 @@ fn test_epoch_rewards_period() {
     let instruction = instruction::get_minimum_delegation();
     env.process_success(&instruction);
 }
+
+// other than Withdraw, instructions should fail with a zero-length stake account
+// account data is truncated for instructions that intend to deallocate the account
+// we want to ensure topping up such an account mid-transaction does not allow reuse
+#[test]
+fn test_no_use_dealloc() {
+    let mut env = Env::init();
+
+    for declaration in &*INSTRUCTION_DECLARATIONS {
+        let is_withdraw = matches!(declaration, StakeInterface::Withdraw { .. });
+        let mut instruction = declaration.to_instruction(&mut env);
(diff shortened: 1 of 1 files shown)


2501babe:
this might also be a good time to think about how we are going to do fuzzing, because it is a small simple change that slihgtly changes behavior. ideally we have some kind of runner via mollusk that tells us something changed between the two versions, we can say "ok this is fine" and adjust the things we expect to be different, and confirm nothing unexpected changed by accident. i have no idea yet what impling this kind of flow would be like since it depends on multiple versions of the program and idk if it can really be part of ci... idk think about it

joncinque (commented):
Looks great to me! I took another looks through the stakes cache, and I believe this will be safe.

The stakes cache only stores stake accounts with delegations, and if lamports go to 0 OR if the account can no longer be deserialized as a stake account, it gets removed from the cache. So essentially we'll go from deserializing `StakeState::Uninitialized` and then failing to get its delegation.

We should be aware that this does create a DOS opportunity on stake accounts. If someone closes a stake account and then tops it up with lamports in the same transaction, it will be impossible to ever reuse it. We could relax the logic in `withdraw` to treat a zero-length account the same as a `StakeState::Uninitialized` -- what do you think?

joncinque (commented):
The new commit looks great to me! And yeah definitely good to think about all of the other interactions

PR #83, opened 2025-08-13
tests: remove native stake test cases

2501babe:
all clusters are running the bpf stake program, so we can stop testing native stake. native stake is in the process of being removed from the agave repo but we actually can remove these tests now,since they exist in agave and will not be removed from there until native stake is removed!

this will let us continue with #82 because we no longer assert behavior of native and bpf stake agree


joncinque (approved):
Looks good to me!

PR #88, opened 2025-10-02
program: bump deps to SDK 3.0

buffalojoec:
Does the whole "busting up" of `solana-program` and `solana-sdk` wherever necessary, upgrading the program up to use the isolated crates on SDK 3.0.

### DIFF
--- a/program/Cargo.toml
+++ b/program/Cargo.toml
@@ -12,33 +12,43 @@ bpf-entrypoint = []
 test-sbf = []
 
 [dependencies]
-arrayref = "0.3.8"
 bincode = "1.3.3"
-borsh = { version = "1.5.1", features = ["derive", "unstable__schema"] }
-num-derive = "0.4"
-num-traits = "0.2"
-num_enum = "0.7.3"
-solana-program = "2.2.1"
-thiserror = "1.0.63"
+solana-account-info = { version = "3.0.0", features = ["bincode"] }
+solana-clock = "3.0.0"
+solana-cpi = "3.0.0"
+solana-instruction-error = "2.0.0"
+solana-msg = "3.0.0"
+solana-native-token = "3.0.0"
+solana-program-entrypoint = "3.0.0"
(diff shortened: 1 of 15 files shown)


program/src/lib.rs
```diff
@@ -1,14 +1,12 @@
-use solana_program::native_token::LAMPORTS_PER_SOL;
+use solana_native_token::LAMPORTS_PER_SOL;
 
 pub mod helpers;
 pub mod processor;
 
 #[cfg(feature = "bpf-entrypoint")]
 pub mod entrypoint;
 
-pub use solana_program;
```
joncinque, line 9: We could eventually re-export some core types, but I haven't found that super useful in the past, so I'm happy to have this removal

program/Cargo.toml
```diff
@@ -12,33 +12,43 @@ bpf-entrypoint = []
...
-solana-program-runtime = "2.2.0"
-solana-stake-interface = { version = "1", features = ["bincode"] }
-solana-system-interface = { version = "1", features = ["bincode"] }
-solana-vote-program = "2.2.0"
-solana-sdk = "2.2.1"
-solana-sdk-ids = "2.2.1"
-solana-sysvar = { version = "2.2.1", features = ["bincode"] }
 rand = "0.8.5"
+solana-account = { version = "3.0.0", features = ["bincode"] }
+solana-config-interface = { version = "1", features = ["serde"] }
```
joncinque, line 38: It looks like the build is ok, but this should probably be v2 to be consistent

joncinque (commented):
Just a tiny thing on the config interface, the rest looks great!

program/src/lib.rs
buffalojoec: Sounds good to me! ✂️

program/Cargo.toml
buffalojoec: Done!

PR #99, opened 2025-10-15
Enable additional clippy rules

grod220:
Aligns clippy checks with [token-2022 makefile](https://github.com/solana-program/token-2022/blob/main/Makefile#L33-L42).

Required excluding generated code due to codama `arithmetic_side_effects` flags and making small syntax & spacing fixes.


rustopian (approved):
Spot on.

PR #152, opened 2025-10-31
Convert warmup/cooldown rates to integers

grod220:
This pull request implements [SIMD-0391](https://github.com/solana-foundation/solana-improvement-documents/pull/391), replacing all floating-point (`f64`) arithmetic in the Stake Program's warmup and cooldown logic with a fixed-point implementation using integer arithmetic. See SIMD for background rationale.

### What's new?

- The warmup/cooldown rate calculation now uses BPS values and u128 integer arithmetic.
- All logic related to this now in isolated `warmup_cooldown_allowance` module
- `Delegation` struct has been updated to maintain backward compatibility with on-chain data. The `warmup_cooldown_rate: f64` field is replaced with `_reserved: [u8; 8]` (preserves memory layout)

### DIFF
--- a/interface/src/state.rs
+++ b/interface/src/state.rs
@@ -29,15 +29,18 @@ pub type StakeActivationStatus = StakeHistoryEntry;
 // Means that no more than RATE of current effective stake may be added or subtracted per
 // epoch.
 #[deprecated(
-    since = "3.1.0",
-    note = "Use ORIGINAL_WARMUP_COOLDOWN_RATE_BPS instead"
+    since = "3.2.0",
+    note = "Use `warmup_cooldown_allowance::ORIGINAL_WARMUP_COOLDOWN_RATE_BPS` instead"
 )]
 pub const DEFAULT_WARMUP_COOLDOWN_RATE: f64 = 0.25;
-#[deprecated(since = "3.1.0", note = "Use TOWER_WARMUP_COOLDOWN_RATE_BPS instead")]
+#[deprecated(
+    since = "3.2.0",
+    note = "Use `warmup_cooldown_allowance::TOWER_WARMUP_COOLDOWN_RATE_BPS` instead"
+)]
 pub const NEW_WARMUP_COOLDOWN_RATE: f64 = 0.09;
 pub const DEFAULT_SLASH_PENALTY: u8 = ((5 * u8::MAX as usize) / 100) as u8;
 
-#[deprecated(since = "3.1.0", note = "Use warmup_cooldown_rate_bps() instead")]
+#[deprecated(since = "3.2.0", note = "Use warmup_cooldown_rate_bps() instead")]
 pub fn warmup_cooldown_rate(current_epoch: Epoch, new_rate_activation_epoch: Option<Epoch>) -> f64 {
     if current_epoch < new_rate_activation_epoch.unwrap_or(u64::MAX) {
         DEFAULT_WARMUP_COOLDOWN_RATE
@@ -534,7 +537,7 @@ impl Delegation {
 
     /// Previous implementation that uses floats under the hood to calculate warmup/cooldown
     /// rate-limiting. New `stake_v2()` uses integers (upstream eBPF-compatible).
-    #[deprecated(since = "3.1.0", note = "Use stake_v2() instead")]
+    #[deprecated(since = "3.2.0", note = "Use stake_v2() instead")]
     pub fn stake<T: StakeHistoryGetEntry>(
         &self,
         epoch: Epoch,
@@ -548,7 +551,7 @@ impl Delegation {
     /// Previous implementation that uses floats under the hood to calculate warmup/cooldown
     /// rate-limiting. New `stake_activating_and_deactivating_v2()` uses integers (upstream eBPF-compatible).
     #[deprecated(
-        since = "3.1.0",
+        since = "3.2.0",
         note = "Use stake_activating_and_deactivating_v2() instead"
     )]
     pub fn stake_activating_and_deactivating<T: StakeHistoryGetEntry>(
@@ -638,7 +641,7 @@ impl Delegation {
     }
 
     // returned tuple is (effective, activating) stake
-    #[deprecated(since = "3.1.0", note = "Use stake_and_activating_v2() instead")]
+    #[deprecated(since = "3.2.0", note = "Use stake_and_activating_v2() instead")]
     fn stake_and_activating<T: StakeHistoryGetEntry>(
         &self,
         target_epoch: Epoch,
@@ -940,7 +943,7 @@ pub struct Stake {
 }
 
 impl Stake {
-    #[deprecated(since = "3.1.0", note = "Use stake_v2() instead")]
+    #[deprecated(since = "3.2.0", note = "Use stake_v2() instead")]
     pub fn stake<T: StakeHistoryGetEntry>(
         &self,
         epoch: Epoch,
--- a/interface/src/state.rs
+++ b/interface/src/state.rs
@@ -1988,117 +1988,6 @@ mod tests {
         check_flag(StakeFlags::empty(), 0);
     }
 
-    #[cfg(test)]
-    #[allow(deprecated)]
-    mod delegation_prop_tests {
-        use {
-            super::*,
-            crate::{
-                stake_history::{StakeHistory, StakeHistoryEntry},
-                ulp::max_ulp_tolerance,
-            },
-            proptest::prelude::*,
-            solana_pubkey::Pubkey,
-        };
-
-        prop_compose! {
-            fn arbitrary_delegation()(
-                // This tests is bounded to the range where `f64` can represent every integer exactly.
-                // Beyond this, integer math and float math diverge considerably.
-                // This case is covered in `warmup_cooldown_allowance.rs`.
-                stake in 0u64..=(1u64 << 53) - 1,
-                activation_epoch in 0u64..=50,
-                deactivation_offset in 0u64..=50,
-            ) -> Delegation {
-                let deactivation_epoch = activation_epoch.saturating_add(deactivation_offset);
-
-                Delegation {
-                    voter_pubkey: Pubkey::new_unique(),
-                    stake,
-                    activation_epoch,
-                    deactivation_epoch,
-                    ..Delegation::default()
-                }
-            }
-        }
-
-        prop_compose! {
-            fn arbitrary_stake_history(max_epoch: Epoch)(
-                entries in prop::collection::vec(
-                    (
-                        0u64..=max_epoch,
-                        0u64..=1_000_000_000_000, // effective
-                        0u64..=1_000_000_000_000, // activating
-                        0u64..=1_000_000_000_000, // deactivating
-                    ),
-                    0..=((max_epoch + 1) as usize),
-                )
-            ) -> StakeHistory {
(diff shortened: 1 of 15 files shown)


interface/src/state.rs
```diff
@@ -537,11 +543,13 @@ impl Delegation {
             })
         {
             // target_epoch > self.deactivation_epoch
+            //
+            // We advance epoch-by-epoch from just after the deactivation epoch up to the target_epoch,
+            // removing (cooling down) the account's share of effective stake each epoch,
+            // potentially rate-limited by cluster history.
```
grod220: Code comments + variable name change below were to add clarity in areas I had some trouble mentally grasping

program/tests/interface.rs
```diff
@@ -99,10 +97,6 @@ const CUSTODIAN_RIGHT: Pubkey =
 // while also making it easy to write tests involving partial (de)activations
 // if the warmup/cooldown rate changes, this number must be adjusted
 const PERSISTENT_ACTIVE_STAKE: u64 = 100 * LAMPORTS_PER_SOL;
-#[test]
-fn assert_warmup_cooldown_rate() {
-    assert_eq!(warmup_cooldown_rate(0, Some(0)), NEW_WARMUP_COOLDOWN_RATE);
-}
```
grod220: Coverage in `interface/src/warmup_cooldown_allowance.rs`

program/Cargo.toml
```diff
@@ -19,7 +19,9 @@ solana-program-entrypoint = "3.0.0"
 solana-program-error = "3.0.0"
 solana-pubkey = "3.0.0"
 solana-rent = "3.0.0"
-solana-stake-interface = { version = "2", features = ["bincode", "borsh", "sysvar"] }
+solana-stake-interface = { path = "../interface", version = "2", features = ["bincode", "borsh", "sysvar"] }
+# TODO: Remove when mollusk updates to use latest interface package
+solana-stake-interface-legacy = { package = "solana-stake-interface", version = "2", features = ["bincode", "borsh", "sysvar"] }
```
grod220: Noticed that the program did not have a local dependency on the interface crate. Feels a bit easier to do local dev by doing this. However, this did trip up a dep inconsistency with mollusk that would need a temp alias here.

2501babe: @joncinque maybe could comment on this, we used to use a local dependency but it was changed to a crates io dependency with fcd8b926 and i dont know if there was a reason or not

joncinque (commented):
Looks good overall! Mostly nits and a few suggestions

interface/src/state.rs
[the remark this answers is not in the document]
grod220: What do you think about crate aliasing on the agave side?

```rust
[dependencies]
solana-stake-interface = "3.0"
solana-stake-interface-legacy = { package = "solana-stake-interface", version = "=2.0.1" }

use solana_stake_interface::{stake_activating_and_deactivating as stake_v3};
use solana_stake_interface_legacy::{
      stake_activating_and_deactivating as stake_legacy,
};

fn feature_gate(gate: &FeatureSet) {
    if gate.is_active() {
        stake_v3()?
    } else {
        stake_legacy()?
    }
}

```

program/Cargo.toml
grod220: Oh, in this case using a patch is definitely better than package aliasing. That way external packages depending on the interface crate will default to the local version as well. Will add.

interface/src/warmup_cooldown_allowance.rs
```diff
@@ -0,0 +1,418 @@
...
+                // We only assert that our implementation correctly falls back to account_portion.
+                prop_assert_eq!(integer_math_result, account_portion);
+            } else {
+                prop_assert!(integer_math_result <= account_portion);
+                prop_assert!(float_math_result <= account_portion);
+                // The `f64`-based oracle calculation loses precision when `u64` inputs exceed 2^53.
+                // The new integer-based implementation is more precise using `u128` arithmetic.
+                // The small difference seen in failing tests is a measure of this legacy inaccuracy.
+                let diff = integer_math_result.abs_diff(float_math_result) as u128;
+                prop_assert!(diff <= 3750, "candidate={}, oracle={}, diff={}", integer_math_result, float_math_result, diff);
```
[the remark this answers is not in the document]
grod220: This is a manual choice based on a close approximation of the max imprecision of the floats. However, there is a more exact way to calculate this. Will update.

interface/src/state.rs
```diff
@@ -674,6 +698,189 @@ impl Delegation {
...
+                    current_epoch,
+                    remaining_deactivating_stake,
+                    &prev_cluster_stake,
+                    new_rate_activation_epoch,
+                );
+
+                // Substract the newly deactivated stake, clamping the per-epoch decrease to at
+                // least 1 lamport so cooldown always makes progress
+                remaining_deactivating_stake =
+                    remaining_deactivating_stake.saturating_sub(newly_deactivated_stake.max(1));
```
joncinque: Confirming this is the only difference in this new function

joncinque (commented):
This is looking really good! The proptests are excellent, just some bits about the deprecation strategy

joncinque (dismissed):
This looks ready to me! Feel free to do any other testing you want

interface/src/warmup_cooldown_allowance.rs
```diff
@@ -200,24 +196,46 @@ mod test {
...
+        // The ideal result (with infinite precision) is 25% of the stake.
+        // 400M * 0.25 = 100M
+        let ideal_allowance = supply_lamports / 4;
+
+        // With saturation fix:
+        // Numerator saturates to u128::MAX (≈ 3.4e38)
+        // Denominator = 4e17 * 10,000 = 4e21
+        // Result = 3.4e38 / 4e21 ≈ 8.5e16 (85M SOL)
+        // 85M is ~21.25% of the stake (fail-safe)
+        // If we allowed unlocking the full account portion it would have been 100% (fail-open)
```
joncinque: Thanks for spelling this all out, it's really useful to see a concrete example of what could happen.

Can you add these as actual asserts in the test?

interface/src/warmup_cooldown_allowance.rs
```diff
@@ -394,8 +412,8 @@ mod test {
                 prop_assert_eq!(float_math_result, 0);
             } else if would_overflow {
                 // In the u128 overflow region, the `f64` implementation is guaranteed to be imprecise.
-                // We only assert that our implementation correctly falls back to account_portion.
-                prop_assert_eq!(integer_math_result, account_portion);
+                // We only assert that the result does not exceed the account's balance
+                prop_assert!(integer_math_result <= account_portion);
```
joncinque: Is it always strictly less than if it'll overflow? Or is it possible to be equal?

joncinque (commented):
Makes sense to me! Just a nit and a question

joncinque (dismissed):
Looks great to me, thanks for addressing all the comments!

rustopian (dismissed):
Excellent testing approach 🦾

rustopian (approved):
Looks good, just rebased 👍

joncinque (dismissed):
Sorry for the lateness, looks good to me!

grod220:
Blockers to move forward

- [x] https://github.com/solana-foundation/solana-improvement-documents/pull/391
- [x] https://github.com/solana-program/stake/pull/303 released (+ any security remediations)
- [x] Publish latest solana-stake-interface crate
- [x] https://github.com/anza-xyz/mollusk/pull/203
- [x] https://github.com/solana-program/stake/pull/308
- [x] This branch rebased & re-reviewed

After merging
- [ ] Refresh: https://github.com/anza-xyz/agave/pull/9637
- [ ] Firedancer impl

interface/src/state.rs
```diff
@@ -700,6 +714,199 @@ impl Delegation {
             (delegated_stake, 0)
         }
     }
+
```
grod220, line 730: Diff below represents a copy/paste of legacy funcs w/ a few code comment clarifications and usage of the new warmup_cooldown_allowance() fns.

interface/src/warmup_cooldown_allowance.rs
```diff
@@ -0,0 +1,426 @@
...
+pub const TOWER_WARMUP_COOLDOWN_RATE_BPS: u64 = 900; // 9%
+
+#[inline]
+pub fn warmup_cooldown_rate_bps(epoch: Epoch, rate_change_activation_epoch: Option<Epoch>) -> u64 {
+    if rate_change_activation_epoch.is_some_and(|activation| epoch >= activation) {
+        TOWER_WARMUP_COOLDOWN_RATE_BPS
+    } else {
+        ORIGINAL_WARMUP_COOLDOWN_RATE_BPS
+    }
+}
```
grod220, line 14: Note we have a slight divergence here. The previous one did:

```
 if current_epoch < new_rate_activation_epoch.unwrap_or(u64::MAX) {
        DEFAULT_WARMUP_COOLDOWN_RATE
    } else {
        NEW_WARMUP_COOLDOWN_RATE
    }
```

But I noticed that if you passed: 
- current epoch: `u64::MAX`
- rate change epoch: `None`

It would return the new rate despite that `None` param. Yes, it will never occur in prod, but using `is_some_and` fixes that issue.

grod220:
With [SIMD-0391](https://github.com/solana-foundation/solana-improvement-documents/pull/391) approved and dependency issues fixed, we are ready to roll with this PR finally.

@joncinque can you re-review?

interface/src/state.rs
```diff
@@ -25,10 +28,16 @@ pub type StakeActivationStatus = StakeHistoryEntry;
 
 // Means that no more than RATE of current effective stake may be added or subtracted per
 // epoch.
+#[deprecated(
+    since = "3.1.0",
```
joncinque: nit: It looks like we did a publish since this was created, so we should use `3.2.0` as the `since` field everywhere

interface/src/state.rs
```diff
@@ -25,10 +28,16 @@ pub type StakeActivationStatus = StakeHistoryEntry;
...
+    note = "Use ORIGINAL_WARMUP_COOLDOWN_RATE_BPS instead"
```
joncinque: nit, let's specify where to find it
```suggestion
    note = "Use `warmup_cooldown_allowance::ORIGINAL_WARMUP_COOLDOWN_RATE_BPS` instead"
```

interface/src/state.rs
```diff
@@ -25,10 +28,16 @@ pub type StakeActivationStatus = StakeHistoryEntry;
...
+)]
 pub const DEFAULT_WARMUP_COOLDOWN_RATE: f64 = 0.25;
+#[deprecated(since = "3.1.0", note = "Use TOWER_WARMUP_COOLDOWN_RATE_BPS instead")]
```
joncinque: nit: same here, let's specify where to find it
```suggestion
#[deprecated(since = "3.1.0", note = "Use `warmup_cooldown_allowance::TOWER_WARMUP_COOLDOWN_RATE_BPS` instead")]
```

joncinque (dismissed):
Just some nits, looks great to me otherwise!

interface/src/state.rs
grod220: Think this tripped me up as we are sort of trying to guess the next release version. Will update.

joncinque (approved):
Looks great to me!

PR #161, opened 2025-11-04
[Replaced] tests: Mollusk Initialize tests (1/9)

rustopian:
[Replaced]

### DIFF
--- a/program/tests/helpers/context.rs
+++ b/program/tests/helpers/context.rs
@@ -1,0 +1,115 @@
+use {
+    super::{
+        instruction_builders::InstructionExecution,
+        lifecycle::StakeLifecycle,
+        utils::{add_sysvars, STAKE_RENT_EXEMPTION},
+    },
+    mollusk_svm::{result::Check, Mollusk},
+    solana_account::AccountSharedData,
+    solana_instruction::Instruction,
+    solana_pubkey::Pubkey,
+    solana_stake_program::id,
+};
+
+/// Builder for creating stake accounts with customizable parameters
+pub struct StakeAccountBuilder {
+    lifecycle: StakeLifecycle,
+}
+
+impl StakeAccountBuilder {
+    pub fn build(self) -> (Pubkey, AccountSharedData) {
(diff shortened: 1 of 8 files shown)


program/tests/helpers/context.rs
```diff
@@ -0,0 +1,112 @@
...
+    }
+}
+
+/// Consolidated test context for stake account tests
+pub struct StakeTestContext {
+    pub mollusk: Mollusk,
+    pub rent_exempt_reserve: u64,
+    pub staker: Pubkey,
+    pub withdrawer: Pubkey,
+}
```
rustopian: This is extended in a later PR, when the instructions being tested require it

program/tests/helpers/instruction_builders.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// Fail-safe default: when `test_missing_signers` is `None`, runs the missing-signers
+    /// test (`true`). Callers must explicitly opt out with `.test_missing_signers(false)`.
+    pub fn execute(self) -> mollusk_svm::result::InstructionResult {
+        let default_checks = [Check::success()];
+        let checks = match self.checks {
+            Some(c) if !c.is_empty() => c,
+            _ => &default_checks,
+        };
+
+        let test_missing_signers = self.test_missing_signers.unwrap_or(true);
```
rustopian: Assumed `true`, so tests must explicitly opt out with `.test_missing_signers(false)`

Meant to prevent a test from forgetting to test missing signers.

program/tests/helpers/mod.rs
```diff
@@ -0,0 +1,7 @@
+#![allow(clippy::arithmetic_side_effects)]
+#![allow(dead_code)]
```
rustopian: Since test banks compile individually, we'll soon otherwise get `dead_code` warnings for any helpers which are not used in every single one.

rustopian:
Taking back to draft pending a few small updates based on discussions in Rome

PR #179, opened 2025-11-14
tests: add compute convenience function

2501babe:
im not married to doing this in `interface.rs` but it seems very convenient. when i wrote the readme originally i just eyeballed the numbers tbh

semi-related (since i was checking CU usage with this) it seems like easily 80% of our usage is just bincode so i wonder if when wincode is more stable we can just use that to economize with relatively little effort

### DIFF
--- a/program/tests/interface.rs
+++ b/program/tests/interface.rs
@@ -1125,7 +1125,7 @@ fn test_no_use_dealloc() {
     }
 }
 
-// this prints ballpack compute unit costs suitable for insertion in README.md
+// this prints ballpark compute unit costs suitable for insertion in README.md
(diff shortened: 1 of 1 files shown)


program/tests/interface.rs
```diff
@@ -1123,3 +1124,89 @@ fn test_no_use_dealloc() {
         }
     }
 }
+
+// this prints ballpark compute unit costs suitable for insertion in README.md
+// run with `cargo test --test interface show_compute_usage -- --nocapture --ignored`
+#[test]
+#[ignore]
+fn show_compute_usage() {
```
grod220: Wonder if we'll want to break things out into a benchmarking crate one day. Not sure, but think this is cool for now 👍

PR #180, opened 2025-11-14
IDL: update to use codama macros

rustopian:
*Most of the diff is generated code (clients and lockfiles) - CI demands it*

### Problem
Client generation uses a lot of fragile manual scripting.

### Solution
Update to using Codama macros, including the new Codama support for doc fields and tuples as instruction arguments.

### DIFF
--- a/clients/rust/Cargo.toml
+++ b/clients/rust/Cargo.toml
@@ -17,10 +17,10 @@ num-traits = "0.2"
 serde = { version = "1.0", features = ["derive"], optional = true }
 serde_with = { version = "3.15", optional = true }
 solana-account-info = "^3.0"
-solana-cpi = "^3.0"
-solana-instruction = "^3.0"
-solana-program-error = "^3.0"
-solana-pubkey = { version = "^3.0", features = ["borsh", "serde"] }
+solana-cpi = "3.0"
+solana-instruction = "3.0"
+solana-program-error = "3.0"
+solana-pubkey = { version = "3.0", features = ["borsh", "serde"] }
 thiserror = "2.0"
 
 [dev-dependencies]
--- a/interface/src/error.rs
+++ b/interface/src/error.rs
@@ -14,6 +14,7 @@ use {
     derive(serde_derive::Deserialize, serde_derive::Serialize)
 )]
 pub enum StakeError {
+    // 0
     /// Not enough credits to redeem.
     #[cfg_attr(
         feature = "codama",
@@ -21,6 +22,7 @@ pub enum StakeError {
     )]
     NoCreditsToRedeem,
 
+    // 1
     /// Lockup has not yet expired.
     #[cfg_attr(
         feature = "codama",
@@ -28,6 +30,7 @@ pub enum StakeError {
     )]
     LockupInForce,
 
+    // 2
     /// Stake already deactivated.
     #[cfg_attr(
         feature = "codama",
@@ -35,6 +38,7 @@ pub enum StakeError {
     )]
     AlreadyDeactivated,
 
+    // 3
     /// One re-delegation permitted per epoch.
     #[cfg_attr(
         feature = "codama",
@@ -42,6 +46,7 @@ pub enum StakeError {
     )]
     TooSoonToRedelegate,
 
+    // 4
     /// Split amount is more than is staked.
     #[cfg_attr(
         feature = "codama",
@@ -49,6 +54,7 @@ pub enum StakeError {
     )]
     InsufficientStake,
 
+    // 5
     /// Stake account with transient stake cannot be merged.
     #[cfg_attr(
         feature = "codama",
@@ -56,6 +62,7 @@ pub enum StakeError {
     )]
     MergeTransientStake,
 
+    // 6
     /// Stake account merge failed due to different authority, lockups or state.
     #[cfg_attr(
         feature = "codama",
@@ -65,6 +72,7 @@ pub enum StakeError {
     )]
     MergeMismatch,
 
+    // 7
     /// Custodian address not present.
     #[cfg_attr(
         feature = "codama",
@@ -72,6 +80,7 @@ pub enum StakeError {
     )]
     CustodianMissing,
 
+    // 8
     /// Custodian signature not present.
     #[cfg_attr(
         feature = "codama",
@@ -79,6 +88,7 @@ pub enum StakeError {
     )]
     CustodianSignatureMissing,
 
+    // 9
     /// Insufficient voting activity in the reference vote account.
     #[cfg_attr(
         feature = "codama",
@@ -86,6 +96,7 @@ pub enum StakeError {
     )]
     InsufficientReferenceVotes,
 
+    // 10
     /// Stake account is not delegated to the provided vote account.
     #[cfg_attr(
         feature = "codama",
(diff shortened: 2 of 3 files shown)


codama.json
```diff
@@ -0,0 +1,20 @@
+{
+  "$schema": "https://codama.gg/schemas/codama.json",
+  "idls": [
+    {
+      "path": "interface/idl.json",
+      "format": "anchor"
```
grod220:
```diff
- "format": "anchor"
+ "format": "codama"
```

interface/build.rs
grod220: Relevant feedback from another PR: https://github.com/solana-program/record/pull/130#issuecomment-3562882177. The feedback on build.rs, makefile+ci/cd support, and feature flagging is worth including here too.

interface/src/instruction.rs
```diff
@@ -44,7 +45,9 @@ pub enum StakeInstruction {
     /// [`Authorized`] carries pubkeys that must sign staker transactions
     /// and withdrawer transactions; [`Lockup`] carries information about
     /// withdrawal restrictions.
-    Initialize(Authorized, Lockup),
+    #[codama(account(name = "stake", writable, docs = "Uninitialized stake account"))]
+    #[codama(account(name = "rent_sysvar", docs = "Rent sysvar"))]
+    Initialize { arg0: Authorized, arg1: Lockup },
```
grod220: I think we need codama support for the tuple variant as this is a breaking change

codama.json
```diff
@@ -0,0 +1,20 @@
+{
+  "$schema": "https://codama.gg/schemas/codama.json",
+  "idls": [
+    {
+      "path": "interface/idl.json",
```
grod220: separately: what's the deal with `program/idl.json`? Should that be deleted?

interface/src/state.rs
```diff
@@ -259,17 +262,18 @@ impl StakeStateV2 {
...
 #[cfg_attr(feature = "frozen-abi", derive(solana_frozen_abi_macro::AbiExample))]
 #[cfg_attr(
     feature = "serde",
     derive(serde_derive::Deserialize, serde_derive::Serialize)
 )]
-#[derive(Debug, PartialEq, Eq, Clone, Copy)]
 pub enum StakeAuthorize {
     Staker,
     Withdrawer,
 }
```
grod220: Does this need a codama enum attr?

interface/src/instruction.rs
rustopian: PRed upstream at https://github.com/codama-idl/codama-rs/pull/73

interface/src/bin/generate-idl.rs
rustopian: Nice tip, added all of this

scripts/generate-clients.mts
```diff
@@ -191,10 +116,19 @@ codama.update(
     },
     {
       // Use omitted optional account strategy for all instructions.
+      // + fix discriminator u8 -> u32.
```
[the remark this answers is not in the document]
rustopian: Originally fixed at this level in https://github.com/solana-program/stake/pull/54

This is still the behavior of things. Despite the macro specifying `u32`, the IDL ends up with `u8` and the clients do as well.

If you remove this tiny fix, all clients will have u8s and U8Decoder etc.

interface/src/state.rs
rustopian: Added 🙏

clients/rust/Cargo.toml
```diff
@@ -11,12 +11,16 @@ license-file = "../../LICENSE"
 serde = ["dep:serde", "dep:serde_with"]
 
 [dependencies]
-borsh = "1.5"
+borsh = { version = "1.5", features = ["derive"] }
 num-derive = "0.4"
 num-traits = "0.2"
 serde = { version = "1.0", features = ["derive"], optional = true }
 serde_with = { version = "3.15", optional = true }
-solana-program = "2.2.1"
```
grod220: Nice

interface/src/error.rs
```diff
@@ -1,70 +1,168 @@
...
     feature = "serde",
     derive(serde_derive::Deserialize, serde_derive::Serialize)
 )]
-#[derive(Clone, Debug, PartialEq, Eq)]
 pub enum StakeError {
     // 0
     /// Not enough credits to redeem.
+    #[cfg_attr(
+        feature = "codama",
+        codama(error(code = 0, message = "Not enough credits to redeem"))
```
grod220: I think the error codes are automatically derived from enum position. Aka, can we remove the `code = #`?

program/tests/interface.rs
grod220: Rebase artifacts too

program/tests/stake_instruction.rs
grod220: We probably need these right?

scripts/generate-clients.mts
```diff
@@ -194,7 +119,15 @@ codama.update(
...
         c.assertIsNode(node, 'instructionNode');
-        return { ...node, optionalAccountStrategy: 'omitted' };
+        return {
+          ...node,
+          optionalAccountStrategy: 'omitted',
+          arguments: node.arguments.map((arg) =>
+            arg.name === 'discriminator'
+              ? { ...arg, type: c.numberTypeNode('u32') }
+              : arg
+          ),
```
grod220: Is this due to the codama dep bump where discriminators are added?

grod220 (commented):
Amazing work, think we are close!

interface/src/error.rs
rustopian: They are; we can yank them. Done in 911b925

program/tests/interface.rs
rustopian: These "rebase artifacts" are all https://github.com/solana-program/stake/commit/5cf845b49e515c098966cd495c78af5274abd173. They're on `main` yet shown here; I'll unravel whatever happened

interface/src/instruction.rs
```diff
@@ -94,14 +151,47 @@ pub enum StakeInstruction {
...
+        )),
+        codama(account(name = "withdraw_authority", signer, docs = "Withdraw authority")),
+        codama(account(
+            name = "lockup_authority",
+            optional,
+            signer,
+            docs = "Lockup authority, if before lockup expiration"
+        ))
+    )]
+    // backwards compatibility with old IDL demands old `args` name
```
grod220: Not quite sure what you mean by old IDL, but can we say something like: "Codama requires tuple argument to be named."?

program/tests/stake_instruction.rs
```diff
@@ -1037,6 +1037,7 @@ fn test_stake_checked_instructions() {
 }
 
 #[test]
+#[allow(deprecated)] // TODO: remove need for this
```
grod220: Can you double check these are necessary? I commented them out, ran tests, and didn't see any warnings. If they are, can you add more to the comment on what is tripping it up for the future dev?

interface/src/instruction.rs
rustopian: I'm referring to people currently using the clients and reference `.args`, `.arg0`, etc. directly from the old Anchor IDL (which codama uses until this PR). L

I'd rather have specific field names here, but it's not worth breaking everyone to do so.

I'll rework the comment

rustopian: Comment clarified in ba555d9

PR #188, opened 2025-11-24
Add pinnochio interface + test-utils crates

grod220:
This PR adds introductory crate/test structure and wires it into CI/CD.

New crates:

- `p-stake-test-utils`: downloads a snapshot of all mainnet stake accounts, unpacks it, and exposes `iter_mainnet_stake_accounts_data()` to iterate over account data. This will download only on the first run. This is helpful for the next task of migrating `StakeStateV2` to zerocopy and needing to confirm all accounts deserialize the same in the new implementation.

- `p-stake-interface`: minimal crate with a test that uses the corpus iterator.


rustopian (changes requested):
Requesting some tweaks, in particular to avoid pulling in `reqwest` and all those multiple versions in Cargo.lock

PR #202, opened 2025-12-02
Remove symlink fixture

grod220:
Giving trouble to package publishing: https://github.com/solana-program/stake/actions/runs/19864581104/job/56924196677

The publish job attempts to bundle the file, can't find it, and throws. We don't want the binary bundled anyway. Deleting the symlink resolves the packaging failure.


joncinque (approved):
Nice catch!

PR #220, opened 2025-12-17
tests: Mollusk Initialize tests (1/9)

rustopian:
Problem
`solana-program-test` is on its slow way out.

Solution
Convert these tests to Mollusk, including simple but perfectly equivalent stake tracking.

This part migrates `Initialize` and `InitializeChecked` tests, and only includes helper utilities relevant to these tests. (Some lifecycle management and all stake history management is not yet required and is saved for later PRs.)

Included:

- Add simple `StakeTestContext` pattern
- Migrate `Initialize` tests to mollusk, out of program_test.rs to initialize.rs
- Start relocation of `StakeLifecycle` management to a helper file
- test_case is used to run all `Initialize` tests against `InitializeChecked` as well
- added wrong owner test
- every instruction runs `test_missing_signers` unless this is explicitly set to false

This does NOT migrate tests from the newer stake_instruction.rs, which already uses Mollusk.

### DIFF
--- a/program/tests/initialize.rs
+++ b/program/tests/initialize.rs
@@ -6,12 +6,13 @@ use {
     helpers::{context::StakeTestContext, lifecycle::StakeLifecycle},
     mollusk_svm::result::Check,
     solana_account::{AccountSharedData, ReadableAccount},
+    solana_instruction::Instruction,
     solana_program_error::ProgramError,
     solana_pubkey::Pubkey,
     solana_rent::Rent,
     solana_sdk_ids::{stake::id, system_program::id as system_program_id},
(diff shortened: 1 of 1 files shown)


program/tests/helpers/lifecycle.rs
```diff
@@ -0,0 +1,29 @@
...
+    Activating,
+    Active,
+    Deactivating,
+    Deactive,
+    Closed,
+}
+
+impl StakeLifecycle {
+    /// Create an uninitialized stake account
+    pub fn create_uninitialized_account(self) -> AccountSharedData {
```
rustopian: Stand-in for now; full Lifecycle management is added in next PR #221

PR #221, opened 2025-12-18
tests: Mollusk StakeLifecycle, Deactivate tests (2/9)

rustopian:
Part 2 of 9 in this project.

(diff is cumulative)

### Problem
solana-program-test is on its slow way out.

### Solution
Convert these tests to Mollusk, including simple but perfectly equivalent stake tracking.

This part migrates `Deactivate` tests and updates to full `StakeLifecycle` management. It does not yet add stake tracking when warping across slots; that is added in the next PR.

Included:

- Full `StakeLifecycle` management
- `Deactivate` tests moved from program_test.rs to deactivate.rs
- Some additional common helpers needed by the above
This does NOT migrate the tests from the newer stake_instruction.rs, which already uses Mollusk.

### DIFF
--- a/program/tests/deactivate.rs
+++ b/program/tests/deactivate.rs
@@ -1,0 +1,128 @@
+#![allow(clippy::arithmetic_side_effects)]
+
+mod helpers;
+
+use {
+    helpers::{context::StakeTestContext, lifecycle::StakeLifecycle},
+    mollusk_svm::result::Check,
+    solana_account::ReadableAccount,
+    solana_program_error::ProgramError,
+    solana_pubkey::Pubkey,
+    solana_stake_client::instructions::{DeactivateBuilder, DelegateStakeBuilder},
+    solana_stake_interface::{error::StakeError, state::StakeStateV2},
+    solana_stake_program::id,
+    test_case::test_case,
+};
+
+#[test_case(false; "activating")]
+#[test_case(true; "active")]
+fn test_deactivate(activate: bool) {
+    let mut ctx = StakeTestContext::with_delegation();
+    let min_delegation = ctx.minimum_delegation.unwrap();
+    let rent_exempt_reserve = ctx.rent_exempt_reserve;
+    let staker = ctx.staker;
+    let withdrawer = ctx.withdrawer;
+
+    let (stake, mut stake_account) = ctx
+        .stake_account(StakeLifecycle::Initialized)
+        .staked_amount(min_delegation)
+        .build();
+
+    let (vote, vote_account_data) = ctx.vote_account.clone().unwrap();
+
+    // Deactivating an undelegated account fails
+    ctx.checks(&[Check::err(ProgramError::InvalidAccountData)])
+        .test_missing_signers(false)
+        .execute(
+            DeactivateBuilder::new()
+                .stake(stake)
+                .stake_authority(staker)
+                .instruction(),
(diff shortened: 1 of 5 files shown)


program/tests/helpers/context.rs
```diff
@@ -0,0 +1,238 @@
...
+
+    pub fn with_delegation() -> Self {
+        let mollusk = Mollusk::new(&id(), "solana_stake_program");
+        let minimum_delegation = solana_stake_program::get_minimum_delegation();
+        Self {
+            mollusk,
+            rent_exempt_reserve: STAKE_RENT_EXEMPTION,
+            staker: Pubkey::new_unique(),
+            withdrawer: Pubkey::new_unique(),
+            minimum_delegation: Some(minimum_delegation),
```
grod220: Why not just make this a required field and have it in new()?

program/tests/helpers/utils.rs
```diff
@@ -0,0 +1,76 @@
...
+        Rent::default().minimum_balance(StakeStateV2::size_of()),
+        STAKE_RENT_EXEMPTION
+    );
+}
+
+/// Resolve all accounts for an instruction, including sysvars and instruction accounts
+///
+/// This function re-serializes the stake history sysvar from mollusk.sysvars.stake_history
+/// every time it's called, ensuring that any updates to the stake history are reflected in the accounts.
+pub fn add_sysvars(
```
grod220: I'm on the fence with this helper. Is it better to automatically populate needed accounts or have the test writer enumerate them at the site of the instruction builder? Hmm.

program/tests/deactivate.rs
```diff
@@ -0,0 +1,128 @@
...
+    solana_pubkey::Pubkey,
+    solana_stake_client::instructions::{DeactivateBuilder, DelegateStakeBuilder},
+    solana_stake_interface::{error::StakeError, state::StakeStateV2},
+    solana_stake_program::id,
+    test_case::test_case,
+};
+
+#[test_case(false; "activating")]
+#[test_case(true; "active")]
+fn test_deactivate(activate: bool) {
```
grod220: Not for this PR, but when we later migrate this instruction, I think it would help to revisit these tests and likely break them up into specific cases versus a single test with multiple inside.

program/tests/deactivate.rs
```diff
@@ -0,0 +1,128 @@
...
+    let (vote, vote_account_data) = ctx.vote_account.clone().unwrap();
+
+    // Deactivating an undelegated account fails
+    ctx.checks(&[Check::err(ProgramError::InvalidAccountData)])
+        .test_missing_signers(false)
+        .execute(
+            DeactivateBuilder::new()
+                .stake(stake)
+                .stake_authority(staker)
+                .instruction(),
```
grod220: If we had per-instruction builders, we could essentially could pull all of the work passing ctx defaults into the instructions to be internal to the builder. If we did that, I think we could make individual tests like 50% smaller. Maybe I am still enamored by the token-wrap pattern of having tests like this small:

```rust
#[test]
fn test_incorrect_escrow_owner() {
    let incorrect_escrow_owner = Pubkey::new_unique();
    WrapBuilder::default()
        .unwrapped_escrow_owner(incorrect_escrow_owner)
        .check(Check::err(TokenWrapError::EscrowOwnerMismatch.into()))
        .execute();
}
```

PR #222, opened 2025-12-19
`StakeStakeV2` zero copy api [wincode]

grod220:
Adds a new zero‑copy state API backed by [wincode](https://github.com/anza-xyz/wincode).

### Key Changes

  - Adds `p-stake-interface` crate (`#![no_std]` compatible)
  - Integrates alignment-1 POD types for safe zero-copy from unaligned slices
  - New API via `StakeStateV2`:
    - `StakeStateV2::from_bytes(&[u8]) -> Result<&StakeStateV2, StakeStateError>` for read-only access
    - `StakeStateV2::from_bytes_mut(&mut [u8]) -> Result<&mut StakeStateV2, StakeStateError>` for
  mutations
    - Tag-checked accessors: `meta()`, `stake()`, `meta_mut()`, `stake_mut()`
    - State transitions: `initialize()`, `delegate()`
  - Maintains full ABI compatibility with legacy bincode-encoded stake accounts

### Motivation

Deserialization via bincode/borsh is computationally expensive for programs that only need to access specific fields or verify the state. This zero-copy approach allows for significantly lower compute unit (CU) usage when interacting with stake accounts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7549,9 +7549,9 @@ dependencies = [
 
 [[package]]
 name = "spl-pod"
-version = "0.8.0"
+version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d01b7d2f9167ed4c1ecafd99fdc3ced71bd30f8b8567d707ec0266313fc3019f"
+checksum = "d6f3df240f67bea453d4bc5749761e45436d14b9457ed667e0300555d5c271f3"
 dependencies = [
  "bytemuck",
  "bytemuck_derive",
--- a/pinocchio/interface/Cargo.toml
+++ b/pinocchio/interface/Cargo.toml
@@ -9,7 +9,7 @@ license = { workspace = true }
 
 [dependencies]
 solana-address = { version = "2.2.0", features = ["wincode"] }
-spl-pod = { version = "0.8.0", features = ["wincode"] }
+spl-pod = { version = "0.7.2", features = ["wincode"] }
 wincode = { version = "0.4.4", features = ["derive"] }
 
 [dev-dependencies]
@@ -18,5 +18,4 @@ proptest = "1.10.0"
 solana-address = { version = "2.2.0", features = ["wincode"] }
 solana-pubkey = "3.0.0"
 solana-stake-interface = { path = "../../interface", features = ["bincode"] }
-spl-pod = { version = "0.8.0", features = ["wincode"] }
 test-case = "3.3.1"


pinocchio/interface/src/state/pod.rs
```diff
@@ -0,0 +1,79 @@
...
+        #[derive(Clone, Copy, Debug, PartialEq, Eq, Default, SchemaWrite, SchemaRead)]
+        #[wincode(assert_zero_copy)]
+        pub struct $name(pub [u8; $n]);
+
+        impl $name {
+            pub const fn from_primitive(v: $prim) -> Self {
+                Self(v.to_le_bytes())
+            }
+
+            pub fn get(self) -> $prim {
```
febo:
```suggestion
            #[inline(always)]
            pub const fn get(self) -> $prim {
```

pinocchio/interface/src/state/layout.rs
```diff
@@ -0,0 +1,145 @@
...
+const _: () = assert!(size_of::<Stake>() == 72);
+const _: () = assert!(size_of::<Authorized>() == 64);
+const _: () = assert!(size_of::<Lockup>() == 48);
+const _: () = assert!(size_of::<Delegation>() == 64);
+
+// ======= Compile-time alignment guards =======
+const _: () = assert!(align_of::<StakeStateV2Layout>() == 1);
+const _: () = assert!(align_of::<Meta>() == 1);
+const _: () = assert!(align_of::<Stake>() == 1);
+const _: () = assert!(align_of::<Delegation>() == 1);
```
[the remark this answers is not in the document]
grod220: Removed all, but kept the main size check.

pinocchio/interface/src/state.rs
```diff
@@ -0,0 +1,242 @@
...
+    #[inline]
+    pub(crate) fn from_u32(v: u32) -> Result<Self, StakeStateError> {
+        match v {
+            0 => Ok(Self::Uninitialized),
+            1 => Ok(Self::Initialized),
+            2 => Ok(Self::Stake),
+            3 => Ok(Self::RewardsPool),
+            other => Err(StakeStateError::InvalidTag(other)),
+        }
+    }
```
joncinque: nit: up to you, but it would be more idiomatic rust to impl `TryFrom<u32>` instead of this function

pinocchio/interface/src/state.rs
```diff
@@ -0,0 +1,242 @@
...
+/// 200-byte stake account layout:
+///
+/// ```text
+/// ┌────────┬──────┬────────────┐
+/// │ Offset │ Size │ Field      │
+/// ├────────┼──────┼────────────┤
+/// │   0    │  4   │ Tag        │
+/// │   4    │ 120  │ Meta       │
+/// │  124   │  72  │ Stake      │
+/// │  196   │  1   │ StakeFlags │
```
joncinque: These were never used, so we can probably just go back to calling it padding

pinocchio/interface/src/state.rs
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    /// Returns a mutable reference to `Meta` if in the `Initialized` or `Stake` state.
+    #[inline]
+    pub fn meta_mut(&mut self) -> Result<&mut Meta, StakeStateError> {
+        match self.tag() {
+            StakeStateV2Tag::Initialized | StakeStateV2Tag::Stake => Ok(&mut self.meta),
+            tag => Err(StakeStateError::InvalidStateAccess(tag)),
+        }
+    }
```
joncinque: nit: is there a reason for the mutable ones to return a `Result` and the immutable ones to return an `Option`? Probably better to be consistent

pinocchio/interface/src/state.rs
```diff
@@ -0,0 +1,242 @@
...
+            return Err(StakeStateError::InvalidTransition {
+                from,
+                to: StakeStateV2Tag::Stake,
+            });
+        }
+
+        if from == StakeStateV2Tag::Initialized {
+            self.stake_flags = 0;
+            self.padding.fill(0);
+        }
```
joncinque: Sorry if I'm being dense, but why is this done only for `Initialized`?

pinocchio/interface/tests/helpers/layout.rs
```diff
@@ -14,22 +14,20 @@ use {
 };
 
 pub const TAG_LEN: usize = StakeStateV2Tag::TAG_LEN;
-pub const LAYOUT_LEN: usize = size_of::<StakeStateV2Layout>();
+pub const STATE_LEN: usize = size_of::<StakeStateV2>();
 
 pub const META_OFF: usize = TAG_LEN;
 pub const STAKE_OFF: usize = TAG_LEN + size_of::<Meta>();
 pub const FLAGS_OFF: usize = TAG_LEN + size_of::<Meta>() + size_of::<Stake>();
 pub const PADDING_OFF: usize = FLAGS_OFF + 1;
```
joncinque: nit: doesn't need to be done here, but let's spell out `OFFSET` fully

pinocchio/interface/src/state.rs
```diff
@@ -0,0 +1,242 @@
...
+// compile-time size check
+const _: () = assert!(size_of::<StakeStateV2>() == 200);
+
+impl StakeStateV2 {
+    /// The fixed size of a stake account in bytes.
+    pub const LEN: usize = size_of::<StakeStateV2>();
+
+    /// Parse stake account data into a read-only reference.
+    pub fn from_bytes(data: &[u8]) -> Result<&Self, StakeStateError> {
+        if data.len() < Self::LEN {
```
febo: `<Self as ZeroCopy>::from_bytes`  should validate the length, so maybe we can use `map_err` there instead of duplicating the check.

pinocchio/interface/src/state.rs
```diff
@@ -0,0 +1,242 @@
...
+        Ok(state)
+    }
+
+    /// Parse stake account data into a mutable reference.
+    pub fn from_bytes_mut(data: &mut [u8]) -> Result<&mut Self, StakeStateError> {
+        if data.len() < Self::LEN {
+            return Err(StakeStateError::UnexpectedEof);
+        }
+        let state = <Self as ZeroCopy>::from_bytes_mut(data)?;
+        StakeStateV2Tag::from_u32(state.tag.get())?;
```
febo: Same thing here about adding a comment.

pinocchio/interface/src/error.rs
```diff
@@ -2,18 +2,20 @@ use crate::state::StakeStateV2Tag;
...
     },
-    /// Pass-through for wincode read errors when borrowing layout structs.
+    /// Wincode deserialization error.
     Read(wincode::ReadError),
-    /// Input buffer is shorter than 200 bytes.
+    /// Buffer shorter than 200 bytes.
     UnexpectedEof,
-    /// Pass-through for wincode write errors when serializing layout structs.
+    /// Wincode serialization error.
     Write(wincode::WriteError),
```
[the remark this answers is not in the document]
grod220: True, mapping wincode err now.

pinocchio/interface/src/state.rs
```diff
@@ -0,0 +1,240 @@
...
+    }
+}
+
+impl TryFrom<u32> for StakeStateV2Tag {
+    type Error = StakeStateError;
+
+    #[inline]
+    fn try_from(v: u32) -> Result<Self, Self::Error> {
+        Self::assert_valid_tag(v)?;
+        Ok(unsafe { Self::from_u32_unchecked(v) })
```
febo: We can do the conversion directly here:
```suggestion
        match v {
            0..=3 => Ok(()),
            other => Err(StakeStateError::InvalidTag(other)),
        }
```

PR #241, opened 2026-02-04
update `bytes` for security issue

2501babe:
we presently fail `cargo audit` on this dep

i was going to run `cargo update` but there are issues with getrandom not being able to use integers when building for wasm because it expects a wasm opt-in which wasnt fixed by downgrading all the wasm crates and getrandom (or, the one of the three getrandoms this applies to) itself. this is too boschian for a wednesday night so im ignoring it for now i just want to unblock my pr

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,10 +3,19 @@
 version = 4
 
 [[package]]
-name = "adler2"
-version = "2.0.1"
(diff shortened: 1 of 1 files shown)


joncinque (approved):
Thanks!

PR #270, opened 2026-02-18
interface: Bump SDK crates to new breaking versions

joncinque:
#### Problem

It isn't possible to use the current stake-interface with bumped versions of the SDK in Agave because `SysvarSerialize` was bumped to a new major, and the current interface crate implements v3 of the trait.

#### Summary of changes

`SysvarSerialize` needs to die, but in the short term, unblock the bump by bumping the dependencies here and publishing a new major version.


apfitzge (commented):
lgtm but will defer proper approval to someone who actually works in this repo.

PR #308, opened 2026-03-10
program: depend on v3 interface

grod220:
The interface is on [v3](https://github.com/solana-program/stake/releases/tag/interface%40v3.0.0) but program depends on [v2](https://github.com/solana-program/stake/blob/8b5dd8d95a2d0b44a7965a41bc5b184ed96f897e/program/Cargo.toml#L23). While rebasing https://github.com/solana-program/stake/pull/152 to the latest main, found this to be a blocker.

### DIFF
--- a/program/Cargo.toml
+++ b/program/Cargo.toml
@@ -20,7 +20,7 @@ solana-program-error = "3.0.1"
 solana-pubkey = "4.2.0"
 solana-rent = "3.0.0"
 solana-security-txt = "1.1.2"
-solana-stake-interface = { version = "2", features = ["bincode", "borsh", "sysvar"] }
+solana-stake-interface = { path = "../interface", version = "3.1.0", features = ["bincode", "borsh", "sysvar"] }
 solana-sysvar = "3.0.0"
 solana-sysvar-id = "3.1.0"
 solana-vote-interface = { version = "5.0.0", features = ["bincode"] }
@@ -45,6 +45,9 @@ solana-pubkey = { version = "4.2.0", features = ["rand"] }
 solana-sdk-ids = "3.1.0"
 solana-signature = "3.4.0"
 solana-signer = "3.0.0"
+# TODO: Remove after agave/mollusk/stake-program depend on aligned StakeHistory version
+#       Either solana-stake-interface v3 or new solana-stake-history crate
+solana-stake-interface-v2 = { package = "solana-stake-interface", version = "2.0.1" }
 solana-svm-log-collector = "3.0.0"
 solana-system-interface = { version = "2.0.0", features = ["bincode"] }
 solana-transaction = "3.0.2"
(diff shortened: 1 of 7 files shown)


program/Cargo.toml
```diff
@@ -45,6 +45,9 @@ solana-pubkey = { version = "4.2.0", features = ["rand"] }
 solana-sdk-ids = "3.1.0"
 solana-signature = "3.4.0"
 solana-signer = "3.0.0"
+# TODO: Remove after agave/mollusk/stake-program depend on aligned StakeHistory version
+#       Either solana-stake-interface v3 or new solana-stake-history crate
+solana-stake-interface-v2 = { package = "solana-stake-interface", version = "2.0.1" }
```
grod220, line 50: Points raised https://github.com/anza-xyz/mollusk/pull/228#pullrequestreview-4097630220

StakeHistory sysvar is creating a dependency chain between
```
stake interface -> agave -> mollusk -> stake program
```

Can be resolved by:
1. Agave libraries shipping with updated solana-stake-inteface dependency
2. Or even better, moving solana-stake-history to its own crate.

program/Cargo.toml
```diff
@@ -20,7 +20,7 @@ solana-program-error = "3.0.1"
 solana-pubkey = "4.2.0"
 solana-rent = "3.0.0"
 solana-security-txt = "1.1.2"
-solana-stake-interface = { version = "2", features = ["bincode", "borsh", "sysvar"] }
+solana-stake-interface = { path = "../interface", version = "3.1.0", features = ["bincode", "borsh", "sysvar"] }
```
2501babe, line 23: iirc there was a reason we stopped using `path` in favor of `version` but i can never remember what it was... maybe @joncinque does? or it could have been febo, i dont think it was me

grod220: Oh wait a sec: https://github.com/solana-program/stake/pull/79#discussion_r2263831588

I suppose we want to keep it detached to update the interface & program independently? Why is that desirable though?

program/Cargo.toml
joncinque: We could potentially move stake history to its own crate, since it likely won't change as often.

Note that it wouldn't really help right now since it still depends on `solana-sysvar`, which depends on `solana-account` / `solana-rent`, which is at the root of all this annoyance.

Two versions of a crate is totally fine if there's no other way

program/Cargo.toml
joncinque: The chain of dependencies goes: program -> agave -> interface

This means we can't update agave until the interface is updated, and we can't update the program until agave is updated.

The whole point of separating out the interface was to remove the circular dependency between the program and agave.

As for the dep here, it's probably better to use the local one to help catch issues right away. But it does mean we can also accidentally publish the program crate without the interface and cause issues.

But I don't feel too strongly either way.

joncinque (approved):
Looks great to me!

program/Cargo.toml
grod220: In the worst case, I think someone attempts a stake-program publish when the used stake-interface version isn't published. That seems worthwhile as it requires interface changes to be handled in the program crate as they are changed versus lagging program-sync PRs. 

That said, it's easy enough to drop the `path = "../interface"` if we run into issues later.

program/Cargo.toml
grod220: SysvarSerialize removed from stake & mollusk. Now just agave to go.

PR #394, opened 2026-05-20
feat: Add methods for checked arithmetics to `StakeHistoryEntry`

vadorovsky:
The usage of existing `std::ops::Add` implementation triggers the `arithmetic_side_effects` clippy lint.

To allow addressing it, add the following methods to `StakeHistoryEntry`:

* `checked_add`
* `wrapping_add`
* `saturating_add`

### DIFF
--- a/interface/src/stake_history.rs
+++ b/interface/src/stake_history.rs
@@ -43,6 +43,30 @@ impl StakeHistoryEntry {
             ..Self::default()
         }
     }
+
+    pub fn checked_add(self, rhs: StakeHistoryEntry) -> Option<Self> {
+        Some(Self {
+            effective: self.effective.checked_add(rhs.effective)?,
+            activating: self.activating.checked_add(rhs.activating)?,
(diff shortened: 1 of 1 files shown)


joncinque (approved):
Makes sense to me! We might want to consider deprecating the normal `ops::Add` implementation then

vadorovsky:
> Makes sense to me! We might want to consider deprecating the normal `ops::Add` implementation then

Sadly, Rust doesn't support deprecating specific trait implementations or trait methods. But perhaps we could go for a more drastic measure and remove the implementation entirely?

joncinque:
> Sadly, Rust doesn't support deprecating specific trait implementations or trait methods. But perhaps we could go for a more drastic measure and remove the implementation entirely?

Oh right of course, my mistake :sweat: but yeah I'm fine with doing a full removal during a breaking change. Let's wait on it though.

Maybe we should add a comment to remember to remove it though. What do you think?

vadorovsky:
> Maybe we should add a comment to remember to remove it though. What do you think?

Good idea, #406

PR #397, opened 2026-05-26
Widen ulp tolerance

grod220:
During audit was provided a case where the ULP diff was > 4. Recommended widening to 10:

> With 5 axioms on f64's behavior, I've formally proven, assuming the triple multiply does not overflow and no division by 0 occurs, that a tolerance of 10*ulp is maintained regardless of the account_portion, cluster_effective, or rate_bps in all scenarios.

Practically speaking, this means the integer and float implementations (not including highly unlikely overflow cases) do not differ more than 1280 lamports. 

No impact to product code, just tests.

### DIFF
--- a/interface/src/ulp.rs
+++ b/interface/src/ulp.rs
@@ -36,14 +36,14 @@ pub fn max_ulp_tolerance(candidate: u64, oracle: u64) -> u64 {
     // Get the ULP spacing
     let ulp = ulp_of_u64(mag);
 
-    // Use a 4x ULP tolerance to account for precision error accumulation in the
+    // Use a 10x ULP tolerance to account for precision error accumulation in the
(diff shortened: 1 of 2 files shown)


joncinque (commented):
Makes sense! Can you add a worst-case test where it's possible to get close to the maximum ULP? Or something that fails on 9?

grod220:
> Can you add a worst-case test where it's possible to get close to the maximum ULP?

Was not able to generate an example larger than 5 (proves the current allowance is too small). Added that test.

> Or something that fails on 9?

Anything 10 and lower will now pass. The audit provided a proof it is not possible higher.

joncinque (approved):
Ah ok, I thought they had an example for 9. This works though!

PR #399, opened 2026-05-27
`StakeHistory` size const

grod220:
Adds a public `SIZE` constant for the serialized size of the `StakeHistory` sysvar account (16_392 bytes), so downstream consumers aren't required to have redundant local versions now that `SysvarSerialize` has been dropped for the sysvar.

Related: https://github.com/anza-xyz/agave/pull/12684

### DIFF
--- a/interface/src/stake_history.rs
+++ b/interface/src/stake_history.rs
@@ -7,6 +7,12 @@ use std::ops::Deref;
 
 pub const MAX_ENTRIES: usize = 512; // it should never take as many as 512 epochs to warm up or cool down
 
+/// Serialized size of a single `(Epoch, StakeHistoryEntry)` tuple
+pub(crate) const EPOCH_AND_ENTRY_SERIALIZED_SIZE: usize = 32;
+
+/// Serialized size of `StakeHistory` sysvar account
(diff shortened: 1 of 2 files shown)


interface/src/stake_history.rs
```diff
@@ -7,6 +7,15 @@ use std::ops::Deref;
 
 pub const MAX_ENTRIES: usize = 512; // it should never take as many as 512 epochs to warm up or cool down
 
+/// Serialized size of a single `(Epoch, StakeHistoryEntry)` tuple
+pub(crate) const EPOCH_AND_ENTRY_SERIALIZED_SIZE: usize = 32;
```
buffalojoec, line 11: Up to you, but the const assertion is a nice touch. Wouldn't hurt to add it here as well!

```suggestion
pub(crate) const EPOCH_AND_ENTRY_SERIALIZED_SIZE: usize = 32;
const _: () = assert!(
    EPOCH_AND_ENTRY_SERIALIZED_SIZE
        == std::mem::size_of::<u64>() + std::mem::size_of::<StakeHistoryEntry>()
);
```

(I did my best to fmt this)

buffalojoec (dismissed):
Great! Thanks

buffalojoec (approved):

PR #407, opened 2026-06-02
Update Codama and regenerate clients

lorisleiva:
This PR bumps Codama and its JS/Rust renderers to the latest versions (`codama` 1.6, `@codama/renderers-js` 2.2, `@codama/renderers-rust` 3.1) and regenerates the clients, reconciling the generated output, configuration, and dependencies accordingly.

The renderer major bumps require a few adaptations: the Codama script args now point at the package/crate root rather than the `src/generated` folder; the JS client's tsconfig uses `moduleResolution: "bundler"` so the regenerated code can resolve the `@solana/kit/program-client-core` subpath; and the Rust client now depends on `solana-address` and `spl-collections` (replacing `solana-pubkey` and `kaigan`) with its Solana crates bumped to v3, matching the converged clients.

This client previously exposed an optional `serde` feature on its generated types. The latest renderer replaces `kaigan` with `spl-collections` for size-prefixed collection types, and `spl-collections` does not implement serde, so the generated types can no longer derive `Serialize`/`Deserialize` under a `serde` feature. This PR therefore drops serde support from the client, matching the converged solana-program clients whose generated code carries no serde.


joncinque (approved):
Looks great!

PR #428, opened 2026-06-16
feat(interface): derive StableAbi and StableAbiSample on AbiExample types

kskalski:
Add the `StableAbi` and `StableAbiSample` derives alongside every `AbiExample`
derive in the stake interface, enabling random-sample ABI digesting for
these types.


joncinque (approved):
Looks great!

PR #448, opened 2026-06-26
interface: use new `solana-stake-history` crate

grod220:
Re-exports migrated types from https://github.com/anza-xyz/solana-sdk/pull/766.

Proposed semver bump as **patch** given there are no API changes. Though minor is also defensible given dep graph changes if folks feel otherwise. Publish will follow this PR.

After running
```
cargo semver-checks --manifest-path interface/Cargo.toml
```

There are errs of missing entities:

```rust
solana_stake_interface::sysvar::stake_history::id
solana_stake_interface::sysvar::stake_history::check_id
solana_stake_interface::stake_history::SIZE
solana_stake_interface::stake_history::MAX_ENTRIES
solana_stake_interface::sysvar::stake_history::ID
solana_stake_interface::stake_history::StakeHistoryEntry
solana_stake_interface::state::StakeActivationStatus
solana_stake_interface::sysvar::stake_history::StakeHistorySysvar
solana_stake_interface::stake_history::StakeHistory
solana_stake_interface::stake_history::StakeHistoryGetEntry
```
But all of these are covered in the new re-exports

```rust
pub use solana_stake_history::sysvar::{check_id, id, StakeHistorySysvar, ID};

pub use solana_stake_history::{
    Epoch, StakeHistory, StakeHistoryEntry, StakeHistoryGetEntry, MAX_ENTRIES, SIZE,
};

pub type StakeActivationStatus = StakeHistoryEntry;
```

PR #449, opened 2026-06-29
Add semver check option

grod220:
Utilizes semver option in the UI for crate publshing. Copies description from https://github.com/anza-xyz/solana-sdk/blob/master/.github/workflows/publish-rust.yml#L45-L52.

Relevant after https://github.com/solana-program/stake/pull/448 and a [failed publish](https://github.com/solana-program/stake/actions/runs/28351304959) due to the semver check (which needs disabling for this next run).


joncinque (approved):
Makes sense!

PR #501, opened 2026-08-19
Generate a visited consumer-facing IDL

lorisleiva:
This PR separates the raw interface IDL from the consumer-facing IDL.

## Pipeline

`make generate-idl-interface` now:

1. runs `codama-rs` and writes the macro-generated tree to `interface-idl.json`;
2. runs the Codama CLI's `idl` script, applying all configured `before` visitors;
3. uses `@codama/renderers-core#writeIdlVisitor` to write the resolved tree to `idl.json`.

The Makefile owns both steps, so CI regenerates and checks both artifacts deterministically. `make generate-clients` also runs the `idl`, `js` and `rust` scripts from the same post-visitor tree.

## Result

- `interface-idl.json` is byte-identical to the previously committed raw `idl.json`.
- Consumer-facing `idl.json` now matches the transformed tree used by client generation:
  - `redelegate` removed;
  - `epoch` and `unixTimestamp` aliases resolved (including date-time display);
  - `stakeStateAccount` wrapper present;
  - instruction data flattened/unwrapped;
  - no dangling defined-type links.
- Generated JS/Rust clients remain byte-identical.

Also adds `@codama/renderers-core@^1.4.0` and updates `codama` to `^1.10.2`.

## Verification

- IDL generation is idempotent across repeated runs.
- Builds without Codama; all interface and Codama schema tests pass.
- Rust fmt/clippy clean with warnings denied.
- Both IDLs parse as valid JSON.
- The shared Solana CI setup already installs pnpm dependencies before running the IDL Makefile target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,8 @@ generate-clients:
 
 generate-idl-%:
 	@cargo install --locked --version =0.13.1 codama-cli
-	codama-rs generate-idl $(call make-path,$*) -o idl.json --pretty $(ARGS)
+	codama-rs generate-idl $(call make-path,$*) -o interface-idl.json --pretty $(ARGS)
+	pnpm codama run idl
 
 # Helpers for publishing
 tag-name = $(lastword $(subst /, ,$(call make-path,$1)))
--- a/codama.mjs
+++ b/codama.mjs
@@ -1,7 +1,7 @@
 import * as c from 'codama';
 
 export default {
-    idl: 'idl.json',
+    idl: 'interface-idl.json',
     before: [
         {
             from: 'codama#updateInstructionsVisitor',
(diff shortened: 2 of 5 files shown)


lorisleiva:
@joncinque The TLDR here is: we generate and IDL using Codama Macros but make a few tweaks on the Codama config files before we generate the clients.

Before this PR, the root `idl.json` was the version exported by Codama Macros and therefore was missing all the transforms we manually wrote on the Codama config file.

After this PR, the root `idl.json` is now the full IDL after both Codama Macros generation (intermediary step stored as `interface-idl.json`) _and_ Codama config file transforms.

Let me know if you're happy for the latter to be the IDL consumed by explorers, wallets, etc.

joncinque (dismissed):
Makes sense to me, thanks!

PR #520, opened 2026-08-31
interface: remove sysvars and stake config

### LINKED ISSUE #53: remove all sysvar pubkeys from instructions interface

2501babe:
since v5 went live, the stake program on all networks is equally happy if you give it the sysvar acocunts or do not. so we can stop using these in instruction builders

closes #53

### DIFF
--- a/interface/src/instruction.rs
+++ b/interface/src/instruction.rs
@@ -12,24 +12,9 @@ use {
 };
 #[cfg(feature = "bincode")]
 use {
-    crate::{config, program::ID, state::StakeStateV2},
+    crate::{program::ID, state::StakeStateV2},
     solana_instruction::{AccountMeta, Instruction},
 };
-
-// Inline some constants to avoid dependencies.
-//
-// Note: replace these inline IDs with the corresponding value from
-// `solana_sdk_ids` once the version is updated to 2.2.0.
-
-#[cfg(feature = "bincode")]
-const CLOCK_ID: Pubkey = Pubkey::from_str_const("SysvarC1ock11111111111111111111111111111111");
-
-#[cfg(feature = "bincode")]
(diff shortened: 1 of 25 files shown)


2501babe:
the first commit is the real changes the second commit is just the codama regen

cc @joncinque after this lands the stake program doesnt need sysvars for cpi, so the stake pool can be updated

interface/src/instruction.rs
```diff
@@ -681,8 +572,7 @@ pub enum StakeInstruction {
     ///      plus rent exempt minimum
     ///   1. `[WRITE]` Uninitialized stake account that will hold the redelegated stake
     ///   2. `[]` Vote account to which this stake will be re-delegated
-    ///   3. `[]` Unused account, formerly the stake config
-    ///   4. `[SIGNER]` Stake authority
+    ///   3. `[SIGNER]` Stake authority
     ///
     #[deprecated(since = "2.1.0", note = "Redelegate will not be enabled")]
```
2501babe, line 577: hows that for thoroughness 😆

grod220 (approved):
LGTM. When old api deprecation? 😅

joncinque (commented):
Coming in late, but looks great to me!

PR #532, opened 2026-09-09
`DeactivateDelinquent` for Closed Vote Accounts [SIMD-0608]

grod220:
Implements [SIMD-0608](https://github.com/solana-foundation/solana-improvement-documents/pull/608). The `DeactivateDelinquent` instruction now also accepts stake delegated to a closed vote account, restoring a cleanup path for delegations orphaned in this edge case.

### New Behavior

The delinquency condition is satisfied when the delegated vote account:
- is not owned by the Vote Program (the normal result of closure) or
- is Vote program owned with either a) zero four-byte discriminator, b) all zero data shorter than four bytes, or c) empty data. Only the discriminator space is inspected, so a large zero-filled shell stays within the compute budget.

All other Vote owned data goes through the existing vote-state decoding and five-epoch credit checks.


interface/src/tools.rs
```diff
@@ -44,6 +47,25 @@ fn get_minimum_delegation_return_data() -> Result<u64, ProgramError> {
...
+        &vote_account_info.try_borrow_data()?,
+        vote_state.as_mut(),
+        vote_account_info.key,
+    )
+    .map_err(|_| ProgramError::InvalidAccountData)?;
+    let vote_state = unsafe { Box::from_raw(Box::into_raw(vote_state).cast::<VoteStateV4>()) };
+
+    Ok(vote_state)
+}
+
```
grod220, line 68: Moved from `program/src/processor.rs` so helpers here can use it

program/src/processor.rs
```diff
@@ -1201,39 +1186,31 @@ impl Processor {
...
-                set_stake_state(
-                    stake_account_info,
-                    &StakeStateV2::Stake(meta, stake, stake_flags),
-                )
-            } else {
-                Err(StakeError::MinimumDelinquentEpochsForDeactivationNotMet.into())
-            }
-        } else {
-            Err(ProgramError::InvalidAccountData)
-        }?;
```
grod220, line 1234: Inverted some of this control flow for easier reading

program/tests/stake_instruction.rs
```diff
@@ -446,18 +446,18 @@ fn test_stake_process_instruction() {
...
         ),
         Err(ProgramError::IncorrectProgramId),
     );
     process_instruction_as_one_arg(
         &mollusk,
         &instruction::deactivate_delinquent_stake(
             &Pubkey::new_unique(),
-            &invalid_vote_state_pubkey(),
             &Pubkey::new_unique(),
+            &invalid_vote_state_pubkey(),
```
grod220, line 460: The error order changes because the reference vote account is now checked before the delinquent account in the processor.

interface/src/tools.rs
```diff
@@ -44,6 +47,25 @@ fn get_minimum_delegation_return_data() -> Result<u64, ProgramError> {
...
+    }
+
+    let mut vote_state = Box::new(MaybeUninit::uninit());
+    VoteStateV4::deserialize_into_uninit(
+        &vote_account_info.try_borrow_data()?,
+        vote_state.as_mut(),
+        vote_account_info.key,
+    )
+    .map_err(|_| ProgramError::InvalidAccountData)?;
+    let vote_state = unsafe { Box::from_raw(Box::into_raw(vote_state).cast::<VoteStateV4>()) };
```
joncinque: What's with this change from the original function?

interface/src/tools.rs
```diff
@@ -2,8 +2,11 @@
 use {crate::MINIMUM_DELINQUENT_EPOCHS_FOR_DEACTIVATION, solana_clock::Epoch};
 #[cfg(feature = "bincode")]
 use {
+    solana_account_info::AccountInfo,
     solana_cpi::{get_return_data, invoke_unchecked},
     solana_program_error::ProgramError,
+    solana_vote_interface::state::VoteStateV4,
+    std::mem::MaybeUninit,
```
joncinque: nit: although the crate isn't no-std yet, let's be sure to avoid std when it's easy
```suggestion
    core::mem::MaybeUninit,
```

interface/src/tools.rs
```diff
@@ -69,9 +91,20 @@ pub fn acceptable_reference_epoch_credits(
 
 /// Check if the provided `epoch_credits` demonstrate delinquency over the previous
 /// [`MINIMUM_DELINQUENT_EPOCHS_FOR_DEACTIVATION`].
+#[deprecated(
+    since = "4.5.0",
```
joncinque: Sorry, this version will have to change after my breaking publish 🙏

interface/src/tools.rs
```diff
@@ -87,6 +120,50 @@ pub fn eligible_for_deactivate_delinquent(
...
+    if vote_account_info
+        .try_borrow_data()?
+        .iter()
+        .take(4)
+        .all(|byte| *byte == 0)
+    {
+        return Ok(true);
+    }
+
+    let vote_state = get_vote_state(vote_account_info)?;
```
joncinque, line 159: I think this makes sense, but just to point it out: if `get_vote_state` fails, then it *won't* be possible to use `deactivate_delinquent` on that account.

`get_vote_state` would most likely fail on a very old or very new vote state version, and we would want that to fail loudly.

interface/src/tools.rs
```diff
@@ -156,4 +234,177 @@ mod tests {
...
+                VoteStateVersions::V3(state) => state.epoch_credits = epoch_credits,
+                VoteStateVersions::V4(state) => state.epoch_credits = epoch_credits,
+                VoteStateVersions::Uninitialized => unreachable!(),
+            }
+            Account::new_data_with_space(1, &state, VoteStateV4::size_of(), &vote::id()).unwrap()
+        }
+
+        #[test_case(system_program::id(), vec![], 0; "removed_account")]
+        #[test_case(Pubkey::new_unique(), vec![255], 1; "other_owner_with_invalid_data")]
+        fn test_non_vote_owner(owner: Pubkey, data: Vec<u8>, lamports: u64) {
```
joncinque: nit: the `lamports` field is a bit misleading since it doesn't make a difference in the implementation -- maybe just hardcode it in the test?

interface/Cargo.toml
```diff
@@ -32,6 +33,7 @@ solana-program-error = "3.0.1"
 solana-pubkey = { version = "4.3.0", default-features = false }
 solana-stake-history = "1.0.0"
 solana-system-interface = "3.3.0"
+solana-vote-interface = { version = "5.0.0", features = ["bincode"], optional = true }
```
joncinque: Also my fault, this'll need to be bumped to 7.0.0

program/Cargo.toml
```diff
@@ -51,6 +50,7 @@ solana-stake-interface-v2 = { package = "solana-stake-interface", version = "2.0
 solana-svm-log-collector = "3.0.0"
 solana-system-interface = { version = "3.3.0", features = ["bincode"] }
 solana-transaction = "3.0.2"
+solana-vote-interface = { version = "5.0.0", features = ["bincode"] }
```
joncinque: Same with this, will need to be 7.0.0

program/tests/stake_instruction.rs
```diff
@@ -7713,32 +7714,47 @@ fn test_deactivate_delinquent_incorrect_vote_owner() {
...
-    let (mollusk, instruction_accounts, transaction_accounts) =
+#[test_case(system_program::id(), vec![], 0; "missing")]
+#[test_case(system_program::id(), vec![], 1; "refunded")]
+#[test_case(Pubkey::new_unique(), vec![255], 1; "other_owner")]
+#[test_case(solana_sdk_ids::vote::id(), vec![], 1; "empty")]
+#[test_case(solana_sdk_ids::vote::id(), vec![0; 3], 1; "three_zero_bytes")]
+#[test_case(solana_sdk_ids::vote::id(), vec![0; VoteStateV4::size_of()], 0; "closed_vote_state")]
+#[test_case(solana_sdk_ids::vote::id(), vec![0; VoteStateV4::size_of()], 1; "refunded_vote_state")]
+#[test_case(solana_sdk_ids::vote::id(), vec![0; 10 * 1024 * 1024], 1; "maximum_size_shell")]
+fn test_deactivate_delinquent_closed_vote_account(owner: Pubkey, data: Vec<u8>, lamports: u64) {
```
joncinque: nit: same with this one, the `lamports` field shouldn't be doing anything, so we can hardcode it to be the same for all cases

program/tests/stake_instruction.rs
```diff
@@ -7713,32 +7714,47 @@ fn test_deactivate_delinquent_incorrect_vote_owner() {
...
+    let vote_account = AccountSharedData::from(Account {
+        lamports,
+        data,
+        owner,
+        ..Account::default()
+    });
+    let (mut mollusk, instruction_accounts, transaction_accounts) =
         setup_deactivate_delinquent_test_with_vote_account(vote_account);
+    mollusk.compute_budget.compute_unit_limit = 200_000;
```
joncinque: Just for my own information, why is this needed?

joncinque (commented):
Looks great overall! Just a few small questions

interface/src/tools.rs
grod220: Makes sense 👍

program/tests/stake_instruction.rs
grod220: oh shoot, was draft work. Removed!

interface/src/tools.rs
grod220: The declared rust version in the interface's cargo.toml doesn't support the original syntax. But... I should have just bumped that version tbh. 

It looks like 1.89 is used in the sdk, so will bump to that.

interface/src/tools.rs
```diff
@@ -44,6 +47,25 @@ fn get_minimum_delegation_return_data() -> Result<u64, ProgramError> {
         .map(u64::from_le_bytes)
 }
 
+/// Deserialize a Vote Program-owned account as [`VoteStateV4`].
+#[cfg(feature = "bincode")]
+pub fn get_vote_state(vote_account_info: &AccountInfo) -> Result<Box<VoteStateV4>, ProgramError> {
```
joncinque: since we'll eventually move the stake program to pinocchio, this function will break when we move from `AccountInfo` to `AccountView` -- how about just taking the owner and data bytes separately and avoiding this dependency entirely?

Cargo.lock
```diff
@@ -4432,9 +4432,9 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.23.38"
+version = "0.23.45"
```
joncinque, line 4435: Can we do a separate PR for this? It's tough to track these updates otherwise

joncinque (commented):
Sorry, just a couple more things that I noticed

grod220:
Given my upcoming leave and need to shepherd the feature gate in, moving this to draft until beginning of Nov

PR #556, opened 2026-09-23
interface: add stable-abi feature for digests on stable Rust

kskalski:
#### Problem
- nightly requirement
  - the crate gates its digests on `frozen-abi`
  - `frozen-abi` requires nightly Rust, because its api digester specializes blanket impls
  - the `StableAbi` digester needs no specialization
  - the crate ties `abi_digest` coverage to nightly
- inert forward
  - the crate forwards `solana-instruction/frozen-abi`
  - `solana-instruction` marks that feature unused

#### Summary of Changes
- dependencies
  - bump the abi crates to the releases carrying `stable-abi`
  - take the macro with `default-features = false`, since its default feature selects the api digester
- features
  - add `stable-abi`, forwarding to the dependencies' `stable-abi`
  - enable `stable-abi` from `frozen-abi`, so existing consumers see no change
  - drop the `solana-instruction/frozen-abi` forward, since that feature is inert
- gates
  - move the `StableAbi` derives to `stable-abi`
  - keep the `AbiExample` derives on `frozen-abi`, since the macro's no-op depends on a globally unified feature
  - keep `min_specialization` under `frozen-abi`


joncinque (approved):
Looks great!

PR #557, opened 2026-09-23
interface: support wincode for instructions 

kskalski:
#### Problem
- feature gating
  - the `bincode` feature gates the instruction constructors
  - the same feature gates `tools::get_minimum_delegation`, because that helper calls a constructor
  - a wincode consumer cannot build an instruction without `bincode`
- wire format
  - nothing pins the instruction wire format
  - a divergence between the two codecs passes unnoticed
 
#### Summary of Changes
- derives
  - derive the wincode schemas on `StakeInstruction` and its args structs
  - skip the deprecated types, since they need no new codec
- constructors
  - gate every builder on `bincode` or `wincode`
  - route the builders through `new_instruction`
  - prefer wincode when both features are on, since the bytes match
- digests
  - pin `StakeInstruction` and `StakeStateV2` with `abi_digest` under both serializers
  - enable `stable-abi` in the dev dependency, so `make test-interface` runs the digest tests
- dev dependencies
  - take solana-account with wincode alone
  - its bincode feature adds inherent `new_data` methods, so it would mask `StateMutWincode`
  - drop `bincode` from this crate's own dev feature list, so the tests build the wincode paths
- tests
  - read stake accounts through `StateMutWincode`, since the runtime uses that trait
  - gate the state tests on `serde` and `wincode`, because they need those derives
- docs
  - document the `wincode` feature in the README


joncinque (approved):
Looks great to me!

